PR #34, opened 2023-01-24
Unit Testing/Jest Setup

mhayes853:
This contains the setup I discussed at tonight's meeting. I simply made a directory called `tests` for all test cases, and inside that folder we can modularize different test components as needed. Right now that folder just contains simple test examples with plain functions and a simple react component. The tests can be run with `npm run test`.

I currently don't have any code coverage tracking enabled currently, but if we want to add that I'm down to do so since tracking coverage can help eliminate dead code and/or help identify missing edge cases. However, if used incorrectly as metric (eg. minimum 80% coverage), it can be problematic since the tests themselves should just be written as specifications rather than testing implementation details. Tests that test implementation details naturally break when the implementation of a feature changes, so it's up to us whether or not how we plan to use coverage metrics.

Lastly, this PR doesn't contain a github actions setup, so maybe that could be done next?

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,7 @@
+const config = {
+  verbose: true,
+  preset: "react-native",
+  testEnvironment: "jsdom",
+};
+
+module.exports = config;
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,8 +52,12 @@
       },
       "devDependencies": {
         "@babel/core": "^7.12.9",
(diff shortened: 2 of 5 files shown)


tifapp (approved):

tifapp (commented):
.

nicoletteL (approved):

mhayes853:
Obsoleted by #37

PR #35, opened 2023-01-24
Update README.md

suryatiru:


ZoroarkZark (approved):

seanim0920:
Looks like "amplify init" creates a new app instead of setting up an app on your local.

PR #42, opened 2023-02-03
Updated/Reworked Post Component UI

nicoletteL:
Right now I'm generating the random colors in the post component itself, but this means that the colors will change whenever the page is refreshed/reloaded. One way to fix this would be to generate the color when the post is created and save it as a property which can be done when post functionality is implemented.

Also, I tried excluding red and green, however it doesn't seem to be possible to do so when randomly generating colors. I can try adjusting the limits of red and green in the rgb color spectrum, but it wouldn't guarantee that those colors wouldn't show up. I think one way to limit the colors would be to have a pre-defined list of colors to choose from.


components/common/IconButton.tsx
```diff
@@ -12,6 +12,7 @@ interface Props {
   margin?: number,
   isLabelFirst?: boolean,
   textStyle?: StyleProp<TextStyle>,
+  accessibilityLabel?: string
```
seanim0920, line 15: Good idea to add this

components/headerComponents/NearbyActivities.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
```
seanim0920, line 1: Sorry I forgot to mention earlier, we can remove the scrollview here and use the ListHeaderComponent prop of the FlatList to allow for a sticky header.

hooks/generateRandomColor.js
```diff
@@ -0,0 +1,19 @@
+import {useState} from 'react';
```
seanim0920, line 1: Is this import unused

hooks/generateRandomColor.js
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+   * @param {number} min
+   * @param {number} max
+   */
+    function randomInt(min, max) {
+      return Math.floor(Math.random() * (max - min) ) + min;
```
seanim0920, line 8: Good call out that this color will be randomly generated on every refresh. The easiest solution would be passing in a random seed based on the post’s ID so the color generated for a particular post is the same every time.

package.json
```diff
@@ -11,7 +11,11 @@
   },
   "dependencies": {
     "@aws-amplify/api": "^4.0.63",
+    "@aws-amplify/notifications": "1.0.12",
```
seanim0920, line 14: Add a comment here to explain the new packages you added and what they’re used for

screens/ActivitiesScreen.tsx
```diff
@@ -1,6 +1,8 @@
+import MainStack from "@stacks/MainStack";
```
seanim0920, line 1: Imports seem unused

screens/FeedScreen.tsx
```diff
@@ -306,48 +312,51 @@ export default function FeedScreen({
...
-      processingFunction={getLikedPosts as any}
-      queryOperation={isChallenge ? postsByLikes : postsByChannel}
-      queryOperationName={isChallenge ? "postsByLikes" : "postsByChannel"}
-      filter={{ channel: channel, sortDirection: "DESC" }}
-      renderItem={renderPostItem}
-      keyExtractor={(item: Post) => item.createdAt.toString() + item.userId}
-      onEndReachedThreshold={0.5}
-    />
+    <>
+      <NearbyActivities />
```
seanim0920, line 316: We can exclude this from the PR, I think Matthew had plans to rework the feed screen. You can keep these changes on your local though for testing.

tests/postUI.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+      maxOccupancy={8} hasInvitations={true}
+    />);
+    
+    expect(screen.getByText(props.description)).toBeDefined();
+    expect(screen.getByLabelText('time until')).toBeDefined();
+    expect(screen.getByLabelText('max occupancy')).toBeDefined();
+    expect(screen.getByLabelText('request invitations')).toBeDefined();
+  });
+
+  it("Renders with only time until event starts", () => {
```
seanim0920, line 67: These test cases are super thorough, TY

tests/postUI.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    expect(screen.queryByLabelText('request invitations')).toBeDefined();
+  });
+
+  it("Name is shortened", () => {
+    render(<PostItem item={post} likes={0} reportPost={mockReportPost}
+      writtenByYou={false} operations={mockOps} startTime={time}
+      maxOccupancy={8} hasInvitations={true}
+    />);
+    
+    expect(screen.getByText("Post T.")).toBeDefined();
```
seanim0920, line 109: Is it possible to check if the name is truncated? If so, could we also have a test that checks if a long description is truncated?

tests/postUI.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    render(<PostItem item={post} likes={0} reportPost={mockReportPost}
+      writtenByYou={false} operations={mockOps} startTime={time}
+      maxOccupancy={undefined} hasInvitations={false}
+    />);
+    const timeIcon = screen.queryByLabelText('time icon');
+
+    expect(timeIcon.props.style[0].color).toEqual("red");
+  });
+
+  it("Occupancy becomes red", () => {
```
seanim0920, line 135: It would be nice if we could somehow map each of these test cases to a storybook story, which should make testing easier. What do you think @mhayes853

hooks/generateRandomColor.js
mhayes853: We could also just store the color in hex itself.

mhayes853:
I think one thing to consider here which should simplify things a ton, on an app-wide scale (not limited to just this PR) it seems that a `Post` takes many forms (actual timeline post, chat message, and now possibly one of these events as well). This can overcomplicate things because "actual posts" are fundamentally different from things like chat messages, and I've noticed an interesting amount of properties on the `Post` type that have nothing to do with "actual posts" at all due to this.

I think it would be a lot simpler to think of posts, chat messages, and more recently events as completely separate entities, because they all have different rates of change and capabilities. In this case, I think it makes sense to have a completely separate component just for the events rather than a shared one like we see here. It would not only be slimmer in terms of loc, but it should allow the removal of all of the "Renders with only x" kind of tests. (Also keep in mind I do plan on refactoring the `PostItem` component at some point to make it more compliant with the new project structure.)

components/PostItem.tsx
```diff
@@ -1,44 +1,11 @@
-import API, { graphqlOperation } from "@aws-amplify/api";
 import { APIListOperations } from "@components/APIList";
-import { deletePost, updatePost } from "@graphql/mutations";
-import React, { useRef, useState } from "react";
-import { Alert, Dimensions, FlatList, StyleSheet, Text, TextInput, TouchableOpacity, View } from "react-native";
+import React, { useState, useEffect } from "react";
```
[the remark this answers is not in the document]
nicoletteL: After reviewing all the comments, including Matthew's, I think I understand what to do now. Before I had separated the old code into a new file and re-wrote the component in PostItem, but I think separating it into its own file and own interface would work a lot better.

screens/ActivitiesScreen.tsx
nicoletteL: Thanks, I forgot to remove those.

tests/postUI.test.tsx
seanim0920: That’s a very good point distinguishing between styling and functionality changes, and I agree with it. I was more intending to call out UI edge cases that would be covered by both the storybook and the unit tests. Ex. If we pass in a very long name will it be truncated, will long descriptions be truncated, will the “invite” button be hidden if the event is public, error/loading states, etc. we’d typically add those cases to the storybook but I could also see benefits to turning those into automated tests.

mhayes853: Those examples would be considered functionality. What I mean by styles in particular are minute details like the position of an element, or the grid (padding) between elements, etc. In essence, we don't need to test what every piece of text is for a particular component, only the important pieces based on the functionality.

screens/ActivitiesScreen.tsx
```diff
@@ -43,7 +44,7 @@ const ActivitiesScreen = () => {
         }}
       >
         SandBox to get started
-    </Text>
+      </Text>
```
tifapp, line 47: Could exclude this spacing from PR

package.json
tifapp: ^

seanim0920: ^

hooks/generateRandomColor.js
seanim0920: We can, but that doesn't solve ensuring each post has a unique color. At least for now the color will only be used to identify the post on the map.

components/postComponents/PostHeader.tsx
```diff
@@ -27,10 +27,10 @@ export default function PostHeader({
   const [likes, setLikes] = useState(item.likes);
   const [replies, setReplies] = useState(item.replies);
   const likeDebounce = useRef(false);
-  let incrementLikeSubscription,
-    decrementLikeSubscription,
-    incrementReplySubscription,
-    decrementReplySubscription;
+  let incrementLikeSubscription: { unsubscribe: () => void; },
```
seanim0920, line 30: Thanks for adding types, though the new components are more important to dedicate time to.

components/EventItem.tsx
```diff
@@ -0,0 +1,277 @@
...
+          style={[styles.flexRow, {paddingBottom: '2%'}]}
+        >
+          <ProfileImageAndName
+            textStyle={{
+              fontWeight: writtenByYou ? "bold" : "normal",
+            }}
+            style={styles.profile}
+            userId={item.userId}
+          />
+          <IconButton
```
seanim0920, line 82: The iconbutton component has a "label" prop to add text, and the label can also be styled. It should be a bit easier than adding a text component yourself every time.

components/EventItem.tsx
```diff
@@ -0,0 +1,277 @@
...
+        {/* Bottom Left Icons (time until event, max occupancy) */}
+        <View style={[styles.flexRow, {paddingBottom: '1%'}]}>
+          <View style={[styles.flexRow, {paddingLeft: '2%'}]}>
+            {startTime != null ?
+              <View style={{flexDirection: 'row'}} accessibilityLabel={'time until'}>
+                <IconButton
+                  iconName={"query-builder"}
+                  size={22}
+                  color={isHours ? 'grey' : 'red'}
+                  onPress={() => null}
```
seanim0920, line 115: Using an iconButton for a non-interactive component would still give it a "button press" effect when the user taps it. Maybe we'll need another similar component that's a non-interactive version of the iconbutton, or edit the iconbutton to remove the touchableOpacity when onPress is undefined.

components/PostItem.tsx
```diff
@@ -321,4 +321,4 @@ const styles = StyleSheet.create({
 
     elevation: 1,
   },
-});
+});
```
seanim0920, line 324: Can exclude this spacing change

components/postComponents/CommentsModal.tsx
```diff
@@ -58,17 +58,18 @@ export default function CommentsModal({ item, operations } : Props) {
...
         autoFocus={true}
         channel={SHA256(item.userId + item.createdAt)} //unique id
-        originalParentId={item.createdAt + "#" + item.userId} 
-        footerComponent={undefined} 
-        isFocused={undefined} 
-        style={undefined} 
-        postButtonLabel={undefined} 
-        renderItem={undefined} 
-        onPostAdded={undefined}      
-      />
```
seanim0920, line 71: Thanks, I'll need another PR in the future to fix all these styling issues.

seanim0920 (approved):

components/EventItem.tsx
```diff
@@ -0,0 +1,277 @@
...
+        setTimeUntil(diffHour);
+      }
+    } 
+  }
+
+  useEffect(() => {
+    setTime();
+  }, [startTime]);
+
+  const handleRequestToJoin = () => {
```
tifapp, line 56: I wonder if we should have a variation of the iconbutton that has a built-in toggleable state. But we can consider that later, for now this is very easy to read.

seanim0920, line 56: I wonder if we should have a variation of the iconbutton that has a built-in toggleable state. But we can consider that later, this is currently very easy to read.

ZoroarkZark (approved):

mhayes853 (approved):

PR #43, opened 2023-02-06
Full commit, placing MapComponent in separate file

ZoroarkZark:
As stated in Slack, I have this mapComponent developed in a separate file and I end up transferring the changes over to ActivitiesScreen.tsx when I need to test.
I still need to work on importing this component correctly into the project to be utilised in other parts of the code.
MapComponent.tsx in Components is where the current code is located.
At the moment, it uses Los Angeles as the current location. 
Multiple marker creation has been done, and settings for the map have been developed according to what was suggested on the Trello board.
Currently working on being able to obtain data from an alternate source.


components/MapComponent.tsx
```diff
@@ -0,0 +1,174 @@
...
+  {
+    "key": 2,
+    "place": "Place two",
+    "lat": 34.039761,
+    "lng": -118.256802,
+  },
+]
+}
+
+//IDK PUT THE LOCATION YOU FETCH FROM THE LOCATION SERVICES HERE? Might be different, could just be placeholder
```
tifapp, line 123: @mhayes853 Will device location be included in your dependencies context? We do have a useLocation hook that will give you the device location, but we can worry about implementing that here in the next sprint.

components/MapComponent.tsx
```diff
@@ -0,0 +1,174 @@
...
+  <Marker 
+  key = {report.key}
+  title = {report.place}
+  coordinate = {{ latitude: report.lat, longitude: report.lng}}
+  >
+  </Marker >)
+}
+  
+//Map view component itself
+function MapComponent () {
```
tifapp, line 143: MapComponent could take in the markers as a prop. You can check out StatusIndicator.tsx for an example of how to define props for a component. It would look something like this:

interface Props {
  markers: {
     key: string,
    place: string,
    lat: number,
    lng: number,
   }[];
}

function MapComponent({markers}: Props) {
 ...

components/MapComponent.tsx
```diff
@@ -0,0 +1,174 @@
...
+//IDK PUT THE LOCATION YOU FETCH FROM THE LOCATION SERVICES HERE? Might be different, could just be placeholder
+const area = {
+    latitude: 34.059761,
+    longitude: -118.276802,
+    latitudeDelta: 0.0922,
+    longitudeDelta: 0.0421,
+};
+
+//Function to use map in order to go through objects and use information from them, for marker locations/information
+const mapMarkerCreations = () => {
```
tifapp, line 132: We can move this marker creator function within the MapComponent so the markers can rerender when the data changes.

components/MapComponent.tsx
```diff
@@ -0,0 +1,174 @@
...
+  return state.markers.map((report) => 
+  <Marker
```
tifapp, line 134: We may need to wrap this inner function in a useCallback for performance, but we can come back to that later.

components/MapComponent.tsx
mhayes853: Yeah, we'll probably have some interface called `UserLocationClient` or something along those lines. We can then have a method on the interface which returns an [`AsyncIterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) that we can use to observe the user's location in real time inside say a `useEffect`.

components/MapComponent.tsx
mhayes853: Seems more like a micro-optimization to me. I mostly use `useCallback` and `useMemo` when I need to do some more complex `useEffect` stuff (eg. to mitigate issues related to the dependency array on `useEffect`). Only in a few scenarios have I actually needed the performance boost from `useMemo` which were always with performing some `O(n)+` calculation on a large list (which should be avoided if possible). In this case using `memo` to memoize the entire map marker would probably be a more beneficial optimization if we even *need* it at all.

components/MapComponent.tsx
mhayes853: Those props could also contain things related to enabling camera movement, zooming, etc. if we decide we want a more flexible map for any feature in the near future.

components/MapComponent.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+
+    ------  MAP VIEW ENDS FROM HERE ---------
+
+    </View>
+
+    )
+}
+
+ */
```
mhayes853, line 96: Btw you should probably delete this whole giant comment and replace it with a documentation comment on the `MapComponent`, many of these notes seem more suited to keep on slack or other areas.

app.json
```diff
@@ -23,6 +23,11 @@
       "favicon": "./assets/favicon.png"
     },
     "android": {
+      "config": {
+        "googleMaps": {
+          "apiKey":  //Fill in as needed, no apiKey normally
```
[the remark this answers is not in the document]
ZoroarkZark: Got it!

components/MapComponent.tsx
tifapp: Yes it should go before returning the html tags, sorry I forgot to mention that

components/MapComponent.tsx
tifapp: Good point, since the map markers will only be static icons for now it should be fine.

seanim0920 (approved):

nicoletteL (approved):

mhayes853 (approved):

PR #44, opened 2023-02-06
Reply Notification Deeplinking + More...

mhayes853:
Ok this is quite lengthy (my bad... 😆) so I'll just say what this PR *doesn't* include:

- It currently does not use the `dependencies` branch
- It does not deal with sending the properly formatted notification on the backend
- *This is because I think we need to have more of a discussion on how to write new backend code going forward (automated test plan, vtl vs js/ts, etc.)*
- It does not have a ui that looks like something you would find on [dribble](https://dribbble.com/)
- *Again, this is because we should probably expend UI reformation effort once we have user data* 
- It does not refactor either the `FeedScreen`, `PostItem`, or `CommentsModal`
- *This PR would be even bigger if it did, and it's already big enough...*

Ok so now for what it *did* do:

- Allows the app to open a reply from a notification (the main thing...)
- Gives the app an official url scheme ("tifapp://") (the other main thing...)
- Adds a reply screen (the real main thing...)
- Puts that wonderful `API.graphql` function behind a `GraphQLOperations` interface
- Adds a decent amount of automated tests for all new code
- Introduces a `UserPosts` interface to abstractly fetch the new `UserPost` type
- Introduces a new `UserPost` type which represents an "actual post", and not 10 different things
- Adds a `UserNotifications` interface for dealing with push notifications on the client
- Adds a `toBeDisplayed` jest matcher, which allows for cleaner assertions when writing ui code
- Adds a lot of general test helpers, and hopefully we can build off many of the things here

Anyways, as always many of the things here I'm down to adjust once we get a better feel for how we want to approach things going forward. Also I'll try my best **not** 😆 to open big PRs like this in the future...


mhayes853:
After some thought about it, I think it's probably for the best that we remove `Tagged`. This is because objects are purely references in js/ts, and equality checks are performed referentially. While `Tagged` would provide nice typesafety, it doesn't really work when it comes to equality checks, especially in things like `useEffect` and `useCallback`... So it's probably just better to remove it (and also the `UserPostMap` type which solely existed due to the problem mentioned).

Also there are probably a lot of things around the reply screen which I could've done better (`graphQLUserPosts` could be quite a bit cleaner), but ofc I'm open to other suggestions.

components/PostItem.tsx
```diff
@@ -2,7 +2,16 @@ import API, { graphqlOperation } from "@aws-amplify/api";
 import { APIListOperations } from "@components/APIList";
 import { deletePost, updatePost } from "@graphql/mutations";
 import React, { useRef, useState } from "react";
-import { Alert, Dimensions, FlatList, StyleSheet, Text, TextInput, TouchableOpacity, View } from "react-native";
```
seanim0920, line 5: Looks like more whitespace changes in this file, lmk if I missed anything

components/common/EmptyView.tsx
```diff
@@ -0,0 +1,5 @@
+/**
+ * A view which renders nothing. This is mostly useful for default props
+ * on other components which may optionally show a view.
+ */
+export const EmptyView = () => <></>;
```
seanim0920, line 5: We can also use {null} when we want to render a blank

lib/GraphQLOperations.ts
```diff
@@ -0,0 +1,53 @@
...
+    super(error.message);
+    this.error = error;
+    this.data = data;
+  }
+}
+
+/**
+ * GraphQL operations backed by AWS Amplify.
+ */
+export class AmplifyGraphQLOperations implements GraphQLOperations {
```
seanim0920, line 38: Looks good, ty

tests/MultiplexedError.test.ts
```diff
@@ -0,0 +1,10 @@
+import { MultiplexedError } from "@lib/MultiplexedError";
+
+describe("MultiplexedError tests", () => {
+  it("separates all error messages from the errors with the given separator", () => {
```
seanim0920, line 4: 👍

screens/ActivitiesScreen.tsx
```diff
@@ -6,45 +6,41 @@ const ActivitiesScreen = () => {
   function signOut() {
     const title = "Are you sure you want to sign out?";
     const message = "";
-    Alert.alert(title, message, [
```
seanim0920, line 9: More whitespace changes

lib/posts/PostIDComponents.ts
```diff
@@ -0,0 +1,42 @@
...
+}: PostIDComponents): string => {
+  return `${creationDate.toISOString()}#${userId}`;
+};
+
+/**
+ * Extracts the components from a post id string if able.
+ *
+ * A post id string takes the following form: "{creation date iso formatted}#{user id}".
+ */
+export const componentsFromPostId = (
```
seanim0920, line 28: 👍  Ty

screens/FeedScreen.tsx
```diff
@@ -1,21 +1,21 @@
 //components
-import APIList, { APIListRefType, APIListRenderItemInfo } from "@components/APIList";
```
seanim0920, line 2: Looks like more whitespace changes

lib/posts/UserPosts.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// TODO: - We should probably have a generic dependencies system for this...
+
+const PostsContext = createContext<UserPosts | null>(null);
+
+/**
+ * Returns an instance of `UserPosts` provided by `UserPostsProvider`.
+ */
+export const useUserPostsDependency = () => useContext(PostsContext)!!; // NB: Programmer error if not used under a UserPostsProvider
```
seanim0920, line 145: 👍

screens/HomeScreen.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    if (!viewedReply) return;
+    modalRef.current?.showModal();
+  }, [viewedReply]);
+
+  return (
+    <>
+      {feedView}
+      {viewedReply && replyView && (
+        <Modal ref={modalRef}>
```
seanim0920, line 62: Will the activities screen use a similar structure to render replies?

seanim0920 (approved):

screens/HomeScreen.tsx
mhayes853: Possibly. One thing I think we should keep in mind is that we may have more of these kinds of notifications (eg. a simple one to let you know there's an invite. or someone liked your post), which would require some scenarios where we only need to display a post without a reply.

At that point we should then probably decide on how these screens take in potentially multiple props that are mutually exclusive and can show different notifications. (I think that a similar approach to that `homeScreenViewReplyFromRouteParams` function may not be such a bad idea in that case).

components/common/EmptyView.tsx
mhayes853: If that's more concise then I'm fine with doing that, it's just that in the SwiftUI world (and some other swift libraries), this `EmptyView` notion seems to be preferred because it's more explicit.

nicoletteL (dismissed):

nicoletteL (dismissed):

seanim0920 (approved):

ZoroarkZark (approved):

PR #47, opened 2023-02-09
Updated Dependencies System

mhayes853:
A much more flexible (at the cost of some increased complexity) dependencies system from what we had last time.


lib/dependencies/DependencyValues.ts
```diff
@@ -0,0 +1,160 @@
...
+ * }
+ *
+ * const Component2 = () => {
+ *   // Retrieve multiple dependency values inside a react component.
+ *   const [dependency1, dependency2] = useDependencyValues<[D1Type, D2Type]>([key1, key2]);
+ *   // ...
+ * }
+ * ```
+ *
+ * Whenever a dependency value is accessed with its respective `DependencyKey`, the default
```
seanim0920, line 24: 👍

lib/dependencies/DependencyHooks.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+const DependencyValuesContext = createContext(new DependencyValues())
+
+const useDependencyValuesContext = () => useContext(DependencyValuesContext)
+
+const useUpdateCopiedDependencyValues = (
+  update: (values: DependencyValues) => void
+) => {
+  const currentValues = useDependencyValuesContext()
+  return useMemo(() => {
```
seanim0920, line 234: How does this useMemo work. The value would always be rerendered since the dependencies are objects

seanim0920 (approved):

seanim0920:
Thanks again for writing the docs. Massive W if everything goes smoothly. Since the dependencies are dynamic we'll have to keep in mind the potential of accidentally overwriting/deleting values in the future, and also ensure the system isn't overused as a massive global data store. This certainly has the potential to make development much easier and safer for us.

lib/dependencies/DependencyHooks.tsx
mhayes853: This `useMemo` ensures that we only perform the transformations on the dependencies when we need to (ie. In most cases we only need 1 transform, so why transform on each render unless we need to perform a transform based on a component's state?). 

While yes, the underlying `DependencyValues` object is well... an object, it should still retain the same memory address throughout renders (unless a parent component changes it via another transform) which would ensure that `useMemo` doesn't continuously run.

.lintstagedrc.cjs
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  "*.ts": "eslint --fix",
```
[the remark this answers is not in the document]
mhayes853: Yes

mhayes853:
Agreed, it is meant specifically as a simple context that wraps everything from the "outside world" which is basically the entire data layer, notifications, uuid generation, anything non-deterministic, etc. In this case it makes the whole initialization process easier and limits the need to have 15+ contexts which would have been the alternative. 

If we run into more complicated state management needs it's probably better to bring on something like zustand (just know that I typically dislike *truly* global state and we should try to scope everything if possible).

nicoletteL (approved):

ZoroarkZark (approved):

PR #49, opened 2023-02-15
Bottom Sheet Functionality

nicoletteL:
This PR contains changes to the events UI that was proposed last week (so it doesn't contain any of Matthew's ideas) and revisions to the test for the UI.

The main thing this PR contains though is the bottom sheet functionality for the activities list. There is a bottom sheet at the bottom of the app that can be swipped up and down that includes the activities list. Right now I have mock data rendering in the list so there aren't any calls to the backend.


ZoroarkZark (approved):

suryatiru (dismissed):

nicoletteL:
It turns out that the bottom sheet wasn't being rendered over the screen but inside it, so it didn't render over other components. I changed it from using a bottom sheet to a bottom sheet modal and tweaked its behavior to match how it was when I was using a regular bottom sheet. I also added another snap point, so now there are three places were it snaps (bottom of screen, about 65%, and at the top of the screen).

I also refactored the event code into using an interface as well as the test.

components/EventsList.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const EventsList = () => {
+  eventItems = new GraphQLEventItems()
+  const ids = Array.from(new Array(10), (_, i) => String(i))
+  const events = eventItems.eventsWithIds(ids)
+
+  // hooks
+  const sheetRef = useRef<BottomSheetModal>(null)
+
+  useEffect(() => {
```
seanim0920, line 23: I haven’t seen .present() before, what does it do?

package.json
```diff
@@ -41,7 +42,7 @@
     "expo-haptics": "~11.2.0",
     "expo-image-manipulator": "~10.3.1",
     "expo-image-picker": "~13.1.1",
-    "expo-linking": "^3.3.0",
+    "expo-linking": "~3.1.0",
```
seanim0920, line 45: Is the downgrade intentional

seanim0920 (approved):

ZoroarkZark (approved):

components/EventsList.tsx
nicoletteL: .present() calls opens the modal. By default the modal is closed when the component is rendered, which means it doesn't appear on the screen at all. In order for the user to swipe it up and down the modal has to be open which is why I'm calling .present() when the component is first rendered.

components/EventsList.tsx
```diff
@@ -0,0 +1,53 @@
...
+    sheetRef?.current?.present()
+  }, [])
+
+  // variables
+  const snapPoints = ["4%", "65%", "100%"]
+
+  return (
+    <BottomSheetModalProvider>
```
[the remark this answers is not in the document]
nicoletteL: I'm not too sure, but the documentation for the bottom sheet modal has it wrapped in this provider. I'm guessing it provides some context that the bottom sheet modal needs. The app doesn't load without the modal wrapped in the provider.

jest.config.js
```diff
@@ -2,4 +2,5 @@ module.exports = {
   verbose: true,
   preset: "jest-expo",
   setupFiles: ["<rootDir>/jest/setupNav.js"],
-};
+  setupFilesAfterEnv: ["@testing-library/jest-dom"]
```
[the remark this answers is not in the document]
nicoletteL: I added that when trying a different way to test some of the UI, but it turns out that I didn't need it. I must have forgot to remove that line.

package.json
nicoletteL: When I run expo I get this message: Some dependencies are incompatible with the installed expo package version:
 - expo-linking - expected version: ~3.1.0 - actual version installed: 3.3.0. I downgraded it in order to avoid breaking something

components/EventsList.tsx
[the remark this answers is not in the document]
mhayes853: Looking at the source code of the lib, it basically passes down a function that allows a child component to dismiss the sheet.

components/EventsList.tsx
```diff
@@ -0,0 +1,53 @@
...
+import EventItem from "@components/EventItem"
+import NearbyActivities from "./headerComponents/NearbyActivities"
+import { Events, GraphQLEventItems } from "@lib/events/Events"
+
+let eventItems: Events
+
+const EventsList = () => {
+  eventItems = new GraphQLEventItems()
+  const ids = Array.from(new Array(10), (_, i) => String(i))
+  const events = eventItems.eventsWithIds(ids)
```
mhayes853, line 18: This should use the dependencies system, bc atm this effectively ties GraphQL to the components, which the dependencies system allows you to avoid.

Eg. It should probably look something like this
```tsx
const EventsList = () => {
    const events = useDependencyValue(eventsDependencyKey)
    // ...
}
```

The in the file where the `Events` interface lies.
```ts
// ...
export const eventsDependencyKey = createDependencyKey<Events>(new GraphQLEventItems())
```

But we can worry about that in the near future.

lib/events/Events.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * An interface representing all the collection of all of the posts in the app.
+ */
+export interface Events {
+  /**
+   * Fetches events with the given ids.
+   *
+   * @returns a list of events (right now just mocks a bunch of events)
+   */
+  eventsWithIds: (ids: string[]) => Event[]
```
mhayes853, line 10: Just a heads up that in the near future we'll probably need this to be something like `eventsInLocation` that takes in location coordinates and a radius.

mhayes853 (approved):

components/EventsList.tsx
nicoletteL: Thanks for the example! I'll start to migrate the code to the dependencies system.

PR #50, opened 2023-02-16
Map_refining

ZoroarkZark:
For the most part in this PR, it's meant to make the map component a lot more scalable and able to be modified from outside sources using props and such. 
For now, the data is being stored in another file.
There is a button to re-center the map to the initial location, but I want to try and design it to lock back to the original location. This also uses a custom map style to remove points of interest.

I haven't been using automated unit testing yet in this file, since I have little experience on that, but I would love information on how to do so.
At the moment, I have been having trouble with getting the exact functionality of the map working in Storybook.


nicoletteL (dismissed):

components/MapTestData.tsx
```diff
@@ -0,0 +1,63 @@
+import { StyleSheet } from "react-native";
```
seanim0920, line 1: Move to mocks folder instead of components folder

seanim0920 (approved):

nicoletteL (approved):

components/MapComponent.tsx
```diff
@@ -1,170 +1,127 @@
...
       </MapView>
+      <View
+        style={{
+        position: 'absolute',//use absolute position to show button on top of the map
+        top: '70%', //for center align
+        alignSelf: 'flex-end' //for align to right
+        }}
+      >
+      { 
+        buttonVisible == true &&
```
mhayes853, line 103: `== true` shouldn't be needed when doing this comparison. If you need to check for if the value is simply not undefined you can do `!!buttonVisible && <Component />`.

components/MapComponent.tsx
```diff
@@ -1,170 +1,127 @@
...
+        <Button title="Re-Center" onPress={() => {onRecenter(); setButtonVisible(false);}}/>
```
mhayes853, line 104: One thing we should probably be conscious of in the future should be to not have these debug buttons and states directly in the production components we're working on. Rather we should essentially be able to put the component into a _debug harness_ (which can be a storybook story, or just a normal component if storybook doesn't work for whatever reason) and do our visual testing from the outside. This makes it so that we don't have to come back and remove all of this debug code later on, when the component needs to _actually_ be displayed in the app. Additionally, we're able to keep the visual test harness around so that anyone who needs to play with the UI can do so later in a non-destructive manner.

To give a simple example of this:

```tsx
// BAD - Has debug code inside the component itself
const MapComponent = (props: MapViewProps) => {
    // ...
    const recenterButtonTapped = () => {
        // ...
    }
    return (
        <View>
            {/* We're co-mingling debug logic with stuff from react-native-maps now... not good...  */}
            <MapView {...} />
            <Button title="Re-Center" onPress={recenterButtonTapped} />
       </View>
    )
}

// GOOD - Uses debug code on the outside, and keeps MapComponent separated
const MapPreview = () => {
    const mapRef = useRef(null)
   
    const recenterButtonTapped = () => {
        mapRef.recenterMap()
    }

    return return (
        <View>
            {/* Here we render the MapComponent that we made and not MapView, much better  */}
            <MapComponent ref={mapRef} />
            <Button title="Re-Center" onPress={recenterButtonTapped} />
       </View>
    )
}
```

Also FYI to not copy what I have directly, but just and idea for what this pattern may look like.

mhayes853 (approved):

PR #54, opened 2023-02-21
Event Creation Uitilities/Infra

mhayes853:
This PR contains some essentials to build out event creation and likely other screens in the future, here are the notable changes:

- A standard `DateTimePicker` component was added (which renders slightly different on iOS and Android respectively)
- A `HexColorPicker` was added along with a `HexColor` type
- A dependency key for playing haptics was added
- Some files were moved around into different directories
- The `GraphQLClient` interface now returns a new `Cancellable` type on the `execute` method. `Cancellable` holds a promise, and a cancel function which should cancel the promise.


lib/Color.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * A simple type to ensure that a hex color string takes somewhat the shape
+ * of a hex string.
+ *
+ * While it would be nice to limit the characters in the string to only valid
+ * hex characters, it seems the typescript compiler went on strike when such
```
seanim0920, line 6: Found an open discussion on this https://github.com/microsoft/TypeScript/issues/41160

components/common/IconButton.tsx
```diff
@@ -1,21 +1,27 @@
-import { MaterialIcons } from "@expo/vector-icons";
```
seanim0920, line 1: I thought all the files were linted, ig some were missed

lib/Haptics.ts
```diff
@@ -0,0 +1,93 @@
+import * as ExpoHaptics from "expo-haptics"
+import { createDependencyKey } from "./dependencies"
+
+/**
+ * A haptic event to be played to the user.
```
seanim0920, line 5: Nice, what kind of haptics will we have

lib/MiscDependencyKeys.ts
```diff
@@ -1,6 +1,6 @@
 import { createDependencyKey } from "./dependencies"
 
 /**
- * A dependency key to hold the currently logged in user id.
+ * A `DependencyKey` to hold the currently logged in user id.
```
seanim0920, line 4: I appreciate the attention to detail 😆

lib/events/EventColors.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Colors that can be used for customizing events.
+ */
+export namespace EventColors {
+  export const Turquoise: HexColor = "#26C6A2"
+  export const Red: HexColor = "#FF4949"
+
+  /**
+   * All supported event colors.
+   */
+  export const all = [Red, Turquoise]
```
seanim0920, line 13: Ideas on other colors to add?

lib/GraphQLClient.ts
```diff
@@ -37,19 +38,25 @@ export class GraphQLClientError<T> extends Error {
...
+        const result = (await operation) as GraphQLResult<T>
+        if (result.errors) {
+          throw new GraphQLClientError({
+            data: result.data,
+            errors: result.errors
+          })
+        }
+        return result.data as T
+      })(),
+      cancel: () => API.cancel(operation as Promise<GraphQLResult<T>>)
```
seanim0920, line 58: Didn't realize we could do that. Nice.

lib/posts/UserPosts.ts
```diff
@@ -94,7 +94,7 @@ export class GraphQLUserPosts implements UserPosts {
               createdAt: creationDate.toISOString(),
               userId
             })
-            .then((value: any) => value.getPost)
+            .value.then((value: any) => value.getPost)
```
seanim0920, line 97: Let's make sure nothing else breaks now that we need to call .value of the graphql client.

screens/ActivitiesScreen.tsx
```diff
@@ -47,7 +51,11 @@ const ActivitiesScreen = () => {
           SandBox to get started
         </Text>
       </TouchableOpacity>
-      <EventsList />
+      <HexColorPicker
```
seanim0920, line 54: Move the event creation UI to a separate file or use storybook. We have a PR for that storybook issue btw https://github.com/tifapp/FitnessProject/pull/51

seanim0920 (approved):

lib/Haptics.ts
mhayes853: Rn it just delegates to `expo-haptics`, but on iOS in particular (as long as we support iOS 13+), we can actually add very customized/personalized haptics (which includes adding sound effects) using a framework called [CoreHaptics](https://developer.apple.com/documentation/corehaptics).

Unfortunately, I don't think there's a well maintained react native library that wraps this framework, so we would have to introduce native code if we wanted to use it.

lib/events/EventColors.ts
mhayes853: That will depend on the designs, but we'll probably have the standard Blue, Orange, etc.

lib/posts/UserPosts.ts
mhayes853: Can confirm this to be the case, better that this change got introduced now rather than down the line where we would've had a lot more code using the client.

nicoletteL (approved):

suryatiru (approved):

PR #56, opened 2023-02-27
Extra_mapwork

ZoroarkZark:
For this pull request, I spent some time implementing simple changes from the feedback given to the mapComponent. This PR isn't very big, but these I felt were necessary for allowing other components to utilise it.


components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
-import React, { useState } from "react";
-import { Button, StyleProp, View, ViewStyle } from "react-native";
-import MapView, { Marker, PROVIDER_GOOGLE } from "react-native-maps";
+import { Location } from "lib/location/Location"
+import React, { useState } from "react"
+import { StyleProp, View, ViewStyle } from "react-native"
+import MapView, { Marker, PROVIDER_GOOGLE } from "react-native-maps"
+
+// Type for the map markers, in order to separate their logic out
+export interface mapMarker {
```
mhayes853, line 7: Types should be in pascal case, but we can add a linter rule for this later.

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+        }}
+        pinColor={"blue"}
+        onPress={() => {
+          onMarkerClick(
+            marker.markedMarker.location.latitude,
+            marker.markedMarker.location.longitude
+          )
+          onSelected(marker.markedMarker)
+        }}
+      />
```
mhayes853, line 76: I'm assuming that you're leaving have a custom component for the marker for later. When that happens, consider passing the `onSelected` callback to the child component (which should also perform the `onMarkerClick` in that case) so that we can programatically select the marker in more ways than just a button press.

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
-interface Props {
-  containStyle?: StyleProp<ViewStyle>;
-  mapStyle?: StyleProp<ViewStyle>;
   initialRegion: {
-    latitude: number,
-    longitude: number,
-    latitudeDelta: number,
-    longitudeDelta: number,
+    latitude: number
+    longitude: number
```
mhayes853, line 21: Consider using the `Location` type for the longitude and latitude.

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
-  canScroll: boolean,
-  canZoom: boolean,
-  canRotate: boolean,
-  };
+    canScroll: boolean
+    canZoom: boolean
+    canRotate: boolean
+  }
+
+  extractKey: (event: any) => string
```
mhayes853, line 34: This shouldn't be necessary if you already require a key in the `mapMarker` type.

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+import MapView, { Marker, PROVIDER_GOOGLE } from "react-native-maps"
+
+// Type for the map markers, in order to separate their logic out
+export interface mapMarker {
+  key: string
+  title: string
+  location: Location
+}
+
+interface Props<T> {
```
mhayes853, line 13: This could have a more descriptive name, maybe something like `MapProps` or `MapComponentProps`?

screens/ActivitiesScreen.tsx
```diff
@@ -30,31 +29,13 @@ const ActivitiesScreen = () => {
...
-        options={EventColors.all}
+      <MapComponent
+        containStyle={mapCompStyle.container}
+        mapStyle={mapCompStyle.map}
+        initialRegion={state.initialRegion}
+        markers={state.markers}
+        extractKey={(event) => event.key}
+        movementSettings={state.movementSettings}
       />
     </GestureHandlerRootView>
```
mhayes853, line 40: I know that storybook for some reason was not liking the map, but maybe what you can do instead of dumping everything in the activities screen (which should probably be removed tbh since we have storybook) you could make a `MapPreview` component so that other's who need to mess around with the map can just use that.

mhayes853 (approved):

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+import { Location } from "lib/location/Location"
+import React, { useState } from "react"
+import { StyleProp, View, ViewStyle } from "react-native"
+import MapView, { Marker, PROVIDER_GOOGLE } from "react-native-maps"
+
+// Type for the map markers, in order to separate their logic out
+export interface mapMarker {
+  key: string
+  title: string
+  location: Location
```
seanim0920, line 10: We will likely need a prop to set the marker icon in the future too. Maybe iconURL or something similar

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
   }
-  markers: {
-  key: number,
-  place: string,
-  lat: number,
-  lng: number,
-  pinColor: string
-  }[];
+
+  markers: mapMarker[]
```
seanim0920, line 26: 👍

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+interface Props<T> {
+  // Contain style specifically for style on container for map
+  // Map style specifically for style on map
+  containStyle?: StyleProp<ViewStyle>
+  mapStyle?: StyleProp<ViewStyle>
 
-interface Props {
-  containStyle?: StyleProp<ViewStyle>;
-  mapStyle?: StyleProp<ViewStyle>;
   initialRegion: {
```
seanim0920, line 19: InitialRegion can be ripped straight from the MapProps of the package by importing MapProps from the package and making the type like “MapProps[“initialRegion”]”. Idk if MapProps is the exact name of the type but you get the idea

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+  const markerData = markers.map((markedMarker) => {
+    const key = extractKey(markedMarker)
+    return {
+      markedMarker,
+      isSelected: key === selectedMarker
+    }
+  })
+
+  // Return the markers so that they appear on the map
   const mapMarkerCreations = () => {
```
seanim0920, line 59: Maybe add “array” in the variable name so we know it’s an array

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
-    </Marker >)
+    return markerData.map((marker) => (
+      <Marker
+        key={marker.markedMarker.key}
+        title={marker.markedMarker.title}
+        coordinate={{
+          latitude: marker.markedMarker.location.latitude,
+          longitude: marker.markedMarker.location.longitude
+        }}
+        pinColor={"blue"}
```
seanim0920, line 68: PinColor will also probably need to be added to props

components/MapComponent.tsx
seanim0920: Good idea

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+    ))
   }
 
+  // When clicking on the marker, zoom towards where it is.
   function onMarkerClick (lat: number, long: number) {
     mapRef.current.animateToRegion({
       latitude: lat,
       longitude: long,
       latitudeDelta: initialRegion.latitudeDelta - 0.02,
-      longitudeDelta: initialRegion.longitudeDelta - 0.02,
```
seanim0920, line 57: Pull “0.02” out into it’s own variable so it’s something we can easily subtitute later, maybe depending on screen size or other factors

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
       longitude: initialRegion.longitude,
       latitudeDelta: initialRegion.latitudeDelta,
-      longitudeDelta: initialRegion.longitudeDelta,
+      longitudeDelta: initialRegion.longitudeDelta
     })
+  } */
+
+  // Set the key of the marker that is selected.
+  function onSelected (givenMapMarker: mapMarker) {
+    setSelectedMarker(givenMapMarker.key)
```
seanim0920, line 102: 👍

components/MapComponent.tsx
```diff
@@ -1,123 +1,146 @@
...
+  }
+
+  // Function to give the location of a long press, ideally for a pin place.
+  function onPinPlace (lat: number, lng: number): Location {
+    return {
```
seanim0920, line 107: 👍

screens/ActivitiesScreen.tsx
seanim0920: I’ll try and get it to work with storybook so we won’t need to do that

seanim0920 (approved):

components/MapComponent.tsx
ZoroarkZark: Yeah, I agree; thank you

nicoletteL (approved):

PR #59, opened 2023-02-28
Date Range with Formatting

mhayes853:
Since @nicoletteL is working on the event card ui and I'm doing the event form, we discussed last night that we should format the date ranges the same way for consistency. I've taken that idea a bit further with this PR and I now think that _all_ date ranges in the app should be formatted that way for consistency if possible.

So what I did was made a type called `FixedDateRange` which is a type that guarantees that a start date is always before and end date (with methods to move the dates). There is a convenience function to create a `FixedDateRange` literally called `dateRange` which should save you some keystrokes over doing `new FixedDateRange(start, end)`. This type also contains a `formatted` method which does the formatting

At some point we should probably update the `Event` type with this new type.


lib/date/DateRange.ts
```diff
@@ -0,0 +1,98 @@
...
+      return new FixedDateRange(addSecondsToDate(date, seconds), date)
+    }
+    return new FixedDateRange(this.startDate, date)
+  }
+
+  /**
+   * Formats this date range in a UI friendly way.
+   */
+  formatted () {
+    // TODO: - Should this support multiple locales?
```
seanim0920, line 63: Definitely for the future if we find an expert in translations

seanim0920 (approved):

seanim0920:
This reminds me, the date and time in the server needs to be timezone-agnostic. I wonder if we should cover it in the app or backend.

mhayes853:
I think for now we can keep it simple since we're only dealing with the US, so maybe just UTC times on the backend which get translated accordingly on the frontend.

However, that's probably not a viable solution if we want to expand outside of North America, and we'll then likely need to store much more information alongside the UTC date (timezone rules, etc.)

ZoroarkZark (approved):

nicoletteL (approved):

PR #60, opened 2023-02-28
Reworked Events UI Card

nicoletteL:
This PR contains the reworked UI based off of @mhayes853 designs, adding shadows to the event card and distance button, and updating the UI tests for it.

For the shadows, only the prop elevation works on Android and honestly doesn't look as good as the IOS shadow props. Instead, I used a package called react-native-shadow-2 which creates a Shadow view that renders the same on both Android and IOS (haven't tested with IOS though).

Also, my Expo Go app on my phone must have automatically updated at some point and it no longer supported Expo SDK 45. In order to fix this and be able to test/work on the project I updated expo to 46 and all the dependencies that used it.


components/EventItem.tsx
```diff
@@ -1,326 +1,177 @@
...
-    shadowOpacity: 0.18,
-    shadowRadius: 1.0,
-    elevation: 1
+    paddingHorizontal: "6%",
+    paddingVertical: "3%",
+    borderRadius: 20
   }
 })
+
+export default React.memo(EventItem)
```
mhayes853, line 177: I don't think this memo actually does anything since the event props will just be compared referentially (not to mention the callback functions that we'll likely need to add on to this). We'll need some way to check whether or not the props are _actually_ equal for this memo to do anything substantial.

(Chances are that we probably won't need to memo unless we actually have lag in the app)

mhayes853 (approved):

ZoroarkZark (approved):

.vscode/settings.json
```diff
@@ -13,5 +13,6 @@
   "editor.formatOnType": false,
   "editor.formatOnSave": true,
   "editor.formatOnSaveMode": "file",
-  "files.autoSave": "onFocusChange"
+  "files.autoSave": "onFocusChange",
+  "cSpell.words": ["Placemark", "tinycolor", "UCSC"]
```
seanim0920, line 17: 👍 Didn't know we could actually add words to spellcheck here, that's cool! Maybe we could add some more commonly used terms in the future. Can't think of any atm.

components/EventItem.tsx
```diff
@@ -1,326 +1,176 @@
...
 const EventItem = ({ event }: Props) => {
-  const [requested, setRequested] = useState(false) // If user has requested to join
-  const [numInvitations, setNumInvitations] = useState(0) // Number of requested invitations
-  const [isMin, setIsMin] = useState(false) // If time limit has <> 1 hour left
-  const [isHours, setIsHours] = useState(false) // If time limit has >= 1 hour left
-  const [isDays, setIsDays] = useState(false) // // If time limit has > 24 hours left
-  const [timeUntil, setTimeUntil] = useState(0) // Time (in either days, hours, or minutes) until event starts
-  const [currentCapacity, setCurrentCapacity] = useState(1) // How many users have joined event
-  const CAPACITY_PERCENTAGE = 0.75
+  const numAttendees = 1
```
seanim0920, line 15: Make a note that these will eventually come from the event object.

components/EventItem.tsx
```diff
@@ -1,326 +1,176 @@
...
+      startColor={shadowColor}
+      offset={[0, 3]}
+      style={{ alignSelf: "stretch" }}
+    >
+      <View style={[styles.container]}>
+        {/* Profile Image, Name, More button */}
+        <View style={[styles.topRow, styles.flexRow]}>
+          <Image
+            style={styles.image}
+            source={require("../assets/icon.png")}
```
seanim0920, line 36: We could create a clone of the profile image and name component with this new styling since this will be used a lot throughout the app. @mhayes853 may want to do refactoring on that component first though 😅

components/EventItem.tsx
```diff
@@ -1,326 +1,176 @@
...
 
-  return (
-    <View style={styles.secondaryContainerStyle}>
-      <View style={[styles.nestedReply]}>
-        {/* Header (Event Icon, Event Title, Distance) */}
-        <View style={[styles.flexRow, styles.eventContainerStyle]}>
-          <View style={[styles.spacingTop, { paddingLeft: "3%" }]}>
-            <Icon name="location-pin" color={event.colorHex} />
+          <View style={[styles.location, styles.flexRow]}>
+            <Icon name="location-on" color={event.colorHex} />
```
seanim0920, line 53: These rows could also be extrapolated into their own component similar to the IconButton. Hold off on that for now though until we finalize the designs and know for sure whether we'll reuse this style in the future.

components/EventItem.tsx
```diff
@@ -1,326 +1,176 @@
...
+          <View style={[styles.flexRow, { alignItems: "center" }]}>
+            <Icon name="people-alt" color={event.colorHex} />
+            <Text
+              style={[styles.attendingText, styles.attendingNumber]}
+            >{`${numAttendees}`}</Text>
+            <Text style={styles.attendingText}>{" attending"}</Text>
           </View>
+
+          <Shadow distance={4} startColor={shadowColor} offset={[0, 1]}>
+            <View
```
seanim0920, line 82: Curious if this will work with a larger font size

components/EventsList.tsx
```diff
@@ -50,4 +51,11 @@ const EventsList = () => {
   )
 }
 
+const styles = StyleSheet.create({
+  secondaryContainerStyle: {
+    backgroundColor: "white",
+    padding: "2%"
```
seanim0920, line 57: Absolute values are good for margins/padding

lib/events/Events.ts
```diff
@@ -14,30 +18,50 @@ export interface Events {
...
+        title: "Pickup Basketball",
         repliesCount: 2,
         writtenByYou: true,
-        startTime: date,
-        maxOccupancy: 5,
-        isAcceptingInvitations: true,
-        colorHex: "magenta",
-        distance: 0.5
+        duration: new FixedDateRange(date, date2),
+        colorHex: "#843efa",
```
seanim0920, line 51: If we're only allowing a set amount of colors we may want an enumerated list stored somewhere instead of using the direct hex value. That can wait though.

components/RNUrlPreview.js
```diff
@@ -1,17 +1,21 @@
+/* eslint-disable */
```
seanim0920, line 1: I'll fix this

seanim0920 (approved):

tests/events/eventUI.test.tsx
```diff
@@ -1,78 +1,25 @@
...
-  })
-
-  it("Occupancy becomes red", () => {
-    render(<EventItem event={TestEventItems.mockEvent(time, 1, true, true)} />)
-
-    const occupancyIcon = screen.queryByLabelText(occupancyIconLabel)
-    expect(occupancyIcon?.props.style.color).toEqual("red")
+    const date = screen.getByLabelText("day")
+    console.log(date)
+    expect(screen.getByText("Today 5pm - 7pm")).toBeDefined()
```
mhayes853, line 23: ✅

lib/events/Events.ts
```diff
@@ -14,30 +18,50 @@ export interface Events {
...
+        username: "Nicolette Antisdel",
+        title: "Pickup Basketball",
         repliesCount: 2,
         writtenByYou: true,
-        startTime: date,
-        maxOccupancy: 5,
-        isAcceptingInvitations: true,
-        colorHex: "magenta",
-        distance: 0.5
+        duration: new FixedDateRange(date, date2),
```
mhayes853, line 50: Nit: There is a helper function called `dateRange` which exists to make a `FixedDateRange` in less keystrokes instead of doing `new FixedDateRange(...)`.

mhayes853 (approved):

tests/date/FixedDateRangeFormatting.test.ts
```diff
@@ -2,7 +2,7 @@ import { dateRange, FixedDateRange } from "@lib/date"
 
 describe("FixedDateRangeFormatting tests", () => {
   beforeEach(() => {
-    jest.useFakeTimers().setSystemTime(new Date("2023-02-26T12:00:00"))
+    jest.useFakeTimers("modern").setSystemTime(new Date("2023-02-26T12:00:00"))
```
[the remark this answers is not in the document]
nicoletteL: I was having trouble with it earlier since I was getting errors about the timers not being modern. You're right though, after I got rid of the errors I didn't check if I still needed to put "modern" in there.

lib/events/Events.ts
nicoletteL: Yeah, I used dateRange in Event and must've forgot to change the one in Events.

ZoroarkZark (approved):

PR #62, opened 2023-03-05
Feat: Event Form

mhayes853:
Simply a form for editing events, have a look!
B9EBD6F5-2B8E-4356-ADAB-7A81D178940F

The form itself (and as should future forms as well) is powered by [react-hook-form](https://react-hook-form.com/) since it works really well in contexts and doesn't rerender all child components when a value changes.

Atm, a Stack navigator setup is being used for the screen layout. While atm this technically isn't needed, the plan is to introduce more screens to this form (location search) which are best handled through stack navigation.

The Date picker UI is definitely not strong atm, so I didn't implement the solution I had in mind since that would take a non-trivial amount of time. The main issue I have with it is that it doesn't scale well to larger font sizes, and that (on iOS) have very little control over how the picker is styled.

As for the bottom sheet, it doesn't work in storybook at all it seems... There's a bug on iOS where the sheet won't appear sometimes, and an Android bug where the sheet may end up above the keyboard after being dismissed in rare scenarios... It seems the root of these issues has to do with the backdrop, so we'll have to look more into those later...

Lastly there's now quite a bit of generic components and testing tools that I needed to implement the UI, so that's always a plus...


ZoroarkZark (approved):

nicoletteL (approved):

components/eventForm/TextField.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [value, setValue, valueRef] = useEventFormField(fieldName)
+  const { setFocusedField } = useEventFormContext()
+
+  return (
+    <TextInput
+      maxFontSizeMultiplier={FontScaleFactors.xxxLarge}
+      ref={valueRef}
+      value={value}
+      onChangeText={setValue}
+      onFocus={() => setFocusedField(fieldName)}
```
seanim0920, line 38: Clean and readable

components/formComponents/HexColorPicker.tsx
```diff
@@ -33,86 +29,84 @@ export type HexColorPickerProps = {
...
-  style,
-  createAccessibilityLabel = defaultCreateAccessibilityLabel
-}: HexColorPickerProps) => {
+  style
+}: HexColorPickerProps<T>) => {
   const playHaptics = useDependencyValue(hapticsDependencyKey)
 
-  const colorTapped = (option: HexColor) => {
+  const colorTapped = (option: T) => {
     if (Platform.OS === "ios") playHaptics(HapticEvent.SelectionChanged)
```
seanim0920, line 57: Poor Android users 😆

lib/location/Geocoding.ts
```diff
@@ -1,6 +1,6 @@
 import { Placemark } from "./Placemark"
 import ExpoLocation from "expo-location"
-import { createDependencyKey } from "@lib/dependencies"
+import { createDependencyKey } from "../dependencies"
```
seanim0920, line 3: Looks the same

screens/ActivitiesScreen.tsx
```diff
@@ -1,10 +1,17 @@
...
 import { Auth } from "aws-amplify"
-import EventsList from "@components/EventsList"
 import React from "react"
 import { Alert, Text, TouchableOpacity } from "react-native"
 import { GestureHandlerRootView } from "react-native-gesture-handler"
+import { QueryClient, QueryClientProvider } from "react-query"
+import EventFormScreen from "./EventFormScreen"
+import { EventFormTestScreen } from "./testScreens/EventFormTestScreen"
+
+const queryClient = new QueryClient()
```
seanim0920, line 14: Where is this variable used

lib/location/Location.ts
```diff
@@ -12,8 +17,8 @@ export type Location = {
  * A location object meant for tracking purposes.
  */
 export type TrackedLocation = {
-  location: Location
-  trackingDate: Date
+  readonly coordinate: Location
```
seanim0920, line 20: 👍

components/eventForm/EventFormValues.ts
```diff
@@ -0,0 +1,54 @@
...
+    location: values.locationInfo.coordinates,
+    dateRange: values.dateRange,
+    color: values.color,
+    shouldHideAfterStartDate: values.shouldHideAfterStartDate,
+    radiusMeters: values.radiusMeters
+  } as EditEventInput
+}
+
+export const eventColorOptions = [
+  { color: EventColors.Red, accessibilityLabel: "Red" },
```
seanim0920, line 47: Better to use a function that transforms the original colors variable so we have one source of truth

components/eventForm/EventFormValues.ts
```diff
@@ -0,0 +1,54 @@
...
+  readonly dateRange: FixedDateRange
+  readonly color: EventColor
+  readonly shouldHideAfterStartDate: boolean
+  readonly radiusMeters: number
+}
+
+/**
+ * Converts valid `EventFormValues` into an `EditEventInput` type.
+ */
+export const eventEditInputFromFormValues = (values: EventFormValues) => {
```
seanim0920, line 33: 👍

components/eventForm/LocationBanner.tsx
```diff
@@ -0,0 +1,102 @@
...
+        style={styles.bannerNavigationIcon}
+      />
+    </View>
+  )
+}
+
+const LocationInfoBanner = (locationInfo: EventFormLocationInfo) => {
+  return locationInfo.placemarkInfo
+    ? (
+      <PlacemarkInfoBanner {...locationInfo.placemarkInfo} />
```
seanim0920, line 47: To make sure I understand this correctly, the first state is if there's already an address saved in the event and the second state is if there isn't and we need to reverse geocode the address?

components/eventForm/LocationBanner.tsx
```diff
@@ -0,0 +1,102 @@
...
+    )
+    : (
+      <GeocodedLocationInfoBanner {...locationInfo} />
+    )
+}
+
+const GeocodedLocationInfoBanner = (locationInfo: EventFormLocationInfo) => {
+  const { data: placemark, status } = useReverseGeocodeQuery(
```
seanim0920, line 55: Could we combine these states in the GeocodedLocationInfoBanner, something like const placemark = placemarkFromProps ?? placemarkFromQuery, if placemark === null show the skeleton loader.

components/eventForm/SubmitButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+const useSubmit = () => {
+  const { formState, submit, watch, hasEdited } = useEventFormContext()
+  const updateInput = eventEditInputFromFormValues(watch())
+  const canSubmit = !!updateInput && !formState.isSubmitting && hasEdited
+  if (!canSubmit) return undefined
+  return async () => {
+    try {
+      await submit(updateInput)
+    } catch {
+      // TODO: - Should we just forward the actual error message from the thrown error here?
```
seanim0920, line 55: Good question, coordinate with Nico and Tristan on the kinds of errors you want to receive. Although if we support multiple languages in the future I don't think we could easily translate those errors from the backend.

components/eventForm/EventForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+      >
+        {children}
+      </EventFormContext.Provider>
+    </FormProvider>
+  )
+}
+
+/**
+ * Uses the current state of field in the event form. Calling the setter
+ * function that this hook returns only causes a rerender in the components
```
seanim0920, line 93: 👍 Good consideration

components/eventForm/EventForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+  })
+  const { handleSubmit, formState, setFocus } = formMethods
+  const [currentSection, setCurrentSection] = useState<
+    EventFormSection | undefined
+  >()
+  const focusedField = useRef<keyof EventFormValues>("title")
+
+  useEffect(() => {
+    if (!currentSection) {
+      setFocus(focusedField.current)
```
seanim0920, line 61: 👍

components/eventForm/EventForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+export type EventFormSection = "date" | "color" | "advanced"
+
+/**
+ * Props for `EventForm`.
+ */
+export type EventFormProps = {
+  /**
+   * The initial values to use in the form, updates to this field
+   * will not update the current form values.
```
seanim0920, line 23: Could we make it readonly?

seanim0920 (approved):

package.json
```diff
@@ -96,7 +97,7 @@
   "devDependencies": {
     "@babel/core": "^7.18.6",
     "@testing-library/react-native": "^10.1.1",
-    "@types/jest": "^29.2.6",
+    "@types/jest": "^27.5.2",
```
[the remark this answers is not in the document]
mhayes853: The problem was that the types were mismatched with the current jest version

lib/FontScale.ts
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Upper limits the font scaling factor to this value.
+   */
+  maximumScaleFactor?: number
+}
+
+/**
+ * Returns the current font settings on the user's device.
+ *
+ * Use this when you need to scale an icon or some other UI element
```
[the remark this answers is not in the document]
mhayes853: True

components/eventForm/AdvancedSettings.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Advanced settings (hiding after start date, etc.) view for the event form.
+ */
+export const EventFormAdvancedSettings = () => {
+  const [shouldHide, setShouldHide] = useEventFormField(
+    "shouldHideAfterStartDate"
+  )
+  return (
+    <View style={styles.hideSettingContainer}>
```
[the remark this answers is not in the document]
mhayes853: Possibly, didn't want to prematurely abstract anything since we only have 1 advanced setting atm

components/formComponents/FormLabels.tsx
```diff
@@ -0,0 +1,133 @@
...
+  />
+)
+
+const useLabelFontScale = () => {
+  return useFontScale({
+    maximumScaleFactor: FontScaleFactors.xxxLarge
+  })
+}
+
+const useSkeletonHeight = () => {
```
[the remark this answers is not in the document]
mhayes853: That could probably use a better name... I wanted the skeleton view to be as flexible as possible which is why it does its sizing exclusively based on the `style` prop.

components/eventForm/EventFormValues.ts
mhayes853: Ideally yes, but at the same time it would be a little awkward if you consider that it would probably look like this:

```ts
const accessibilityColorMap = {
    // ...
}
export const eventColorOptions = EventColors.all.map((color) => ({ color, accessibilityLabel: accessibilityColorMap[color] ?? "Unknown Color"  }))
```

`"Unknown Color"` would never be returned, so it would seem pretty weird that we should have to handle that case.

components/eventForm/LocationBanner.tsx
mhayes853: Yes

components/eventForm/LocationBanner.tsx
mhayes853: I could possibly do something with disabling the query in react query to make that happen

components/eventForm/SubmitButton.tsx
mhayes853: We wouldn't just use the error from the backend directly, rather we would translate an error code/identifier into an error message on the frontend which would be in the proper language

components/eventForm/EventForm.tsx
mhayes853: There wouldn't really be much of a reason to do so, since you can't really modify props from inside a component...

lib/events/EventColors.ts
```diff
@@ -1,14 +1,29 @@
-import { HexColor } from "@lib/Color"
-
 /**
  * Colors that can be used for customizing events.
  */
 export namespace EventColors {
```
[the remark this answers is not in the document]
mhayes853: Good idea

components/formComponents/DateTimePicker.tsx
```diff
@@ -73,16 +61,6 @@ export type DateTimePickerProps = {
    * hour and minute information.
    */
   formatTime?: (date: Date) => string
-
-  /**
-   * The minimum date for this picker.
-   */
-  minimumDate?: Date
```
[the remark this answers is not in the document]
mhayes853: I've temporarily removed it for now bc it was acting weird on Android and we didn't really need the functionality ASAP. Also I kinda want to redo the UI for the form date time picker, but I didn't implement it fully here since it would take a non-trivial amount of time.

PR #63, opened 2023-03-06
Location Search Result Row UI

mhayes853:
UI that will be displayed inside the location search screen. Here's what the storybook story looks like (I decided to put all the combinations in 1 story bc it would be really annoying to have to switch between 10 different tabs for such a small piece of UI...):
B5A78E8B-FF14-4B50-8F6D-76CAA7A42B50_1_201_a

A couple of things:
- A search result contains the user's distance from the location, but if the distance is really close (< 0.1 miles), the UI simply omits the distance
- A new formatting function was added for formatting the miles, which I also placed inside the event UI component since the search result row and the pill on the event UI card
- The icon switches based on if the location came from the user's _search history_ so that they get clear feedback that the location is one they've used in the past/have used very often


seanim0920 (approved):

ZoroarkZark (approved):

lib/DistanceFormatting.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Formats miles into a compact format suitable for displaying in small UI spaces.
+ *
+ * Format: "{`miles` rounded to the nearest tenth} mi"
+ *
+ * If `miles` < 0.1, then it will use 0.1 as the formatted amount.
+ */
+export const compactFormatMiles = (miles: number) => {
+  // TODO: - Support multiple locales
+  return `${Math.max(parseFloat(miles.toFixed(1)), 0.1)} mi`
```
nicoletteL, line 10: Thanks for adding this!

components/EventItem.tsx
```diff
@@ -89,7 +90,9 @@ const EventItem = ({ event }: Props) => {
               ]}
             >
               <Icon name="near-me" size={20} color="white" />
-              <Text style={styles.distanceText}>{`${distance} mi`}</Text>
+              <Text style={styles.distanceText}>
+                {compactFormatMiles(distance)}
```
nicoletteL, line 94: :white_check_mark:

nicoletteL (approved):

PR #64, opened 2023-03-07
Dynamic_circlecomponent

ZoroarkZark:
This pull request is primarily working on the mapComponent and the pin selection logic. Not only do circles only render if the tapped marker is the one selected, but there is a separate customization array to separate information for the markers on essential and non-essential. At the moment, it works on its own and new test data has been added to make it work. If there is any feedback you can give, let me know!

- Pin dropping logic
- Marker selection logic + circle rendering
- Readjusted zoom based on size of circle given
- Edited map (no longer shows direction popup)


mhayes853 (changes requested):

components/MapComponent.tsx
```diff
@@ -1,19 +1,27 @@
...
-import { StyleProp, View, ViewStyle } from "react-native"
-import MapView, { Marker, PROVIDER_GOOGLE } from "react-native-maps"
+import { StyleProp, StyleSheet, View, ViewStyle } from "react-native"
+import MapView, { Circle, Marker, PROVIDER_GOOGLE } from "react-native-maps"
 
 // Type for the map markers, in order to separate their logic out
-export interface mapMarker {
+export interface MapMarker {
   key: string
   title: string
```
seanim0920, line 10: Color and icon will be set by the user while creating an event, so I'd think they would be included in the event data type @mhayes853 Also I don't think we have color in the event schema yet so ask Nico or Tristan about adding that to the backend.

components/MapComponent.tsx
```diff
@@ -1,19 +1,27 @@
...
   location: Location
 }
 
+export interface MarkerCustomize {
```
seanim0920, line 14: Are there any other style props we can define for the markers? Icon, pin color and circle color should be determined from the event data rather than defined here.

components/MapComponent.tsx
```diff
@@ -77,28 +91,73 @@ export function MapComponent<T extends mapMarker> ({
...
+        color: "gray",
+        circleFillColor: "rgba(0, 0, 0, 0.5)",
+        circleStrokeColor: "gray",
+        circleStrokeWidth: 1
+      }
+    }
+  }
+
+  // When clicking on the marker, zoom towards where it is.
+  function onMarkerClick (lat: number, long: number) {
```
seanim0920, line 139: It looks like the purpose of this function is to center the map on the marker; we can rename the function to something like "centerMapOnMarker()" for clarity.

components/MapComponent.tsx
```diff
@@ -35,11 +48,12 @@ interface Props<T> {
 }
 
 // Map view component itself
-export function MapComponent<T extends mapMarker> ({
+export function MapComponent<T extends MapMarker> ({
   initialRegion,
   markers,
+  customizers,
```
seanim0920, line 54: Not a bad idea to allow each marker to have its own custom style, but passing a customizers array shouldn't be necessary here; the color, icon, etc will be determined from the map marker data.

components/MapComponent.tsx
```diff
@@ -65,7 +79,7 @@ export function MapComponent<T extends mapMarker> ({
           latitude: marker.markedMarker.location.latitude,
           longitude: marker.markedMarker.location.longitude
         }}
-        pinColor={"blue"}
+        pinColor={customizationCreation(marker.markedMarker).color}
```
seanim0920, line 82: Color should be read from the marker data

components/MapComponent.tsx
```diff
@@ -77,28 +91,73 @@ export function MapComponent<T extends mapMarker> ({
...
+          strokeColor={
+            customizationCreation(point.markedMarker).circleStrokeColor
+          }
+          strokeWidth={
+            customizationCreation(point.markedMarker).circleStrokeWidth
+          }
+        />
+      )
+    }
+  }
```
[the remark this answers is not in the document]
seanim0920: If we were to combine the circle jsx with the marker jsx we could do

<>
<Marker {...props} />
{condition === true && <Circle {...props} />}
</>

Alternatively if we want to keep the circle in a separate component we can do
return condition ? <Circle/> : null (or <></>)

A linter rule is a good idea. Nbd at the moment.

seanim0920 (dismissed):

components/MapComponent.tsx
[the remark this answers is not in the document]
seanim0920: We should have a linter for this.

components/MapComponent.tsx
```diff
@@ -23,83 +32,142 @@ interface Props<T> {
...
     canScroll: boolean
     canZoom: boolean
     canRotate: boolean
   }
+}
 
-  extractKey: (event: any) => string
+// When clicking on the marker, zoom towards where it is.
+// Using stackOverflow formula
+// stackoverflow.com/questions/5206018/fit-mapview-to-circle-bounds
```
mhayes853, line 53: The link takes me to some weird 2011 post about setting up reliable UDP sockets in C 😆 (which isn't possible, since UDP is really just the IP protocol in the transport layer)...

components/MapComponent.tsx
```diff
@@ -23,83 +32,142 @@ interface Props<T> {
...
+function circleCoordinateGeneration (lat: number, long: number, radius: number) {
+  const earthRadius = 6378.1 // km
```
mhayes853, line 55: Btw we actually also have a `EARTH_RADIUS_METERS` constant in the same file as `degreesToRadians`, so you could just use that and divide by 1000. It could also remove the small `// km` comment that you have.

components/MapComponent.tsx
```diff
@@ -23,83 +32,142 @@ interface Props<T> {
...
+                  marker.markedMarker.location.latitude,
+                  marker.markedMarker.location.longitude,
+                  initialRadius.radius
+                )
+              }
+            }}
+          />
+          {createdCircle}
+        </>
+      )
```
mhayes853, line 126: We should probably make this more like `FlatList` and call a `renderMarker` prop so that we can use any JSX here which you would call inside the Marker component (and have that `renderMarker` function deal with the circle internally), but we'll worry about it later.

See: https://github.com/react-native-maps/react-native-maps/blob/master/docs/marker.md#children-components

components/MapComponent.tsx
```diff
@@ -23,83 +32,142 @@ interface Props<T> {
...
 
-  markers: mapMarker[]
+  initialRadius: {
+    radius: number
+  }
+
+  markers: MapMarker[]
+  customizers: MarkerCustomize[]
+
+  currentSelectedMarker: String | undefined
```
mhayes853, line 42: ✅

mhayes853 (approved):

components/MapComponent.tsx
```diff
@@ -23,83 +32,142 @@ interface Props<T> {
...
     canZoom: boolean
     canRotate: boolean
   }
+}
 
-  extractKey: (event: any) => string
+// When clicking on the marker, zoom towards where it is.
+// Using stackOverflow formula
+// stackoverflow.com/questions/5206018/fit-mapview-to-circle-bounds
+function circleCoordinateGeneration (lat: number, long: number, radius: number) {
```
mhayes853, line 54: This could also use a better name (and some tests to ensure we don't mess up this somewhat obscure calculation if we ever need to configure it later)

components/MapComponent.tsx
mhayes853: Btw, a conciser way to do this is `currentSelectedMarker?: string`

seanim0920 (approved):

components/MapComponent.tsx
seanim0920: We can consider that in the future if we need to render different kinds of markers besides events, but as of right now I think it's safe to assume all events will have the same style of marker.

nicoletteL (approved):

PR #66, opened 2023-03-07
Event form typography improvements

mhayes853:
Some improvements to the typography of the event form, specifically around the location banner. I also cleaned up the code of the location banner a bit and added a test to ensure that the name and address of the location is displayed when successfully geocoded.

### DIFF
--- a/components/eventForm/LocationBanner.tsx
+++ b/components/eventForm/LocationBanner.tsx
@@ -2,13 +2,11 @@ import { useReverseGeocodeQuery } from "../../hooks/Geocoding"
 import { placemarkToFormattedAddress } from "../../lib/location"
 import React from "react"
 import { StyleSheet, View } from "react-native"
-import {
-  useEventFormContext
-} from "./EventForm"
+import { useEventFormContext } from "./EventForm"
 import { FormLabel, SkeletonFormLabel } from "../formComponents/FormLabels"
 import { MaterialIcons } from "@expo/vector-icons"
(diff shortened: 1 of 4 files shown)


components/eventForm/LocationBanner.tsx
```diff
@@ -41,49 +39,53 @@ export const EventFormLocationBanner = () => {
...
-    style={styles.label}
-    icon="location-pin"
-    headerText={name ?? "Unknown Location"}
-    captionText={address ?? "Unknown Address"}
-  />
-)
+const usePlacemarkInfo = (locationInfo: EventFormLocationInfo) => {
+  const { data: placemark, status } = useReverseGeocodeQuery(
+    locationInfo.coordinates,
+    { enabled: !locationInfo.placemarkInfo }
```
seanim0920, line 74: What state would this apply to?

seanim0920 (approved):

components/eventForm/LocationBanner.tsx
mhayes853: I basically forwarded the `enabled` option from react query, so it won't attempt to do any geocoding if there's already placemark info in the initial form values.

ZoroarkZark (approved):

mhayes853:
Since there are merge conflicts and my plan is to cleanup the form screen anyway in the near future, may as well close this for now and include these changes in a bigger PR that contains more fixes.

PR #67, opened 2023-03-08
Update Event Type

mhayes853:
This PR simply converts the `EventColors` namespace into an enum such that you can now use `Object.values`, `Object.keys`, `Object.entries`, etc. to do transforms on the colors like @seanim0920 suggested. I also updated the `Event` type to use this new enum rather than a string.

### DIFF
--- a/lib/events/Event.ts
+++ b/lib/events/Event.ts
@@ -1,5 +1,6 @@
 import { dateRange, FixedDateRange } from "@lib/date"
 import { Location, Placemark } from "@lib/location"
+import { EventColors } from "./EventColors"
 
 /**
  * A type representing an event hosted by a user, which is meant for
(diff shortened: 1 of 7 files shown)


seanim0920 (approved):

lib/events/Event.ts
```diff
@@ -13,8 +14,8 @@ export type Event = {
   readonly repliesCount: number
   readonly description?: string
   readonly writtenByYou: boolean
-  readonly duration: FixedDateRange
-  readonly colorHex: string
+  readonly dateRange: FixedDateRange
+  readonly color: EventColors
```
nicoletteL, line 18: :+1:

lib/events/EventColors.ts
```diff
@@ -1,29 +1,12 @@
...
  */
-export type EventColor =
-  | typeof EventColors.Red
-  | typeof EventColors.Green
-  | typeof EventColors.Blue
-  | typeof EventColors.Pink
-  | typeof EventColors.Purple
-  | typeof EventColors.Yellow
-  | typeof EventColors.Orange
+export enum EventColors {
```
nicoletteL, line 4: :+1:

nicoletteL (approved):

ZoroarkZark (approved):

PR #68, opened 2023-03-08
Material Icon Component

mhayes853:
We've been using a lot of material icons with the `useFontScale` hook and size 24, might as well create a separate component for it.


seanim0920 (approved):

nicoletteL:
Would it be worth it to make a custom component (or some sort of option) with these settings for material community icons? I'm using those in the frontend as material icons have a very limited selection of outlined icons. I'm not sure if we'll be using anymore outlined icons in the future though.

mhayes853:
Sure, since they're both material icons after all. I think then this should be a proclamation to _only_ use Material Icons in the UI then for consistency.

nicoletteL (approved):

seanim0920 (approved):

PR #71, opened 2023-03-09
Location Search History Persistence

mhayes853:
The location search feature allows for the user to search existing addresses using any arbitrary text. However, the user may have no idea of _what_ to search, or maybe they just want to host an event in the same location they attended last week. And so my proposed solution to that is to store their most recent results on the search screen so they can quickly access them. 

#63 introduced what an individual search result looks like in UI, and one of the things you may have noticed is that there's a different icon on the last result in the UI screenshot which represents a result that comes from the user's search history. Naturally, we need some way to tell if a result is part of the user's search history, and that's what this PR is for... A way to persist location search history using `AsyncStorage`.

### Detailed Design

There's an interface called `LocationSearchHistory` which contains an implementation called `AsyncStorageLocationHistory`. This interface has 3 functions:

1. Querying the entire history ordered by the most recent active history items.
    - Allows for the search screen to display an initial list of results before the user has searched anything.
2. Querying individual history items based on their location.
    - Allows for determining if a result returned from Amplify Geo is a part of the search history.
3. Saving a search result into history.
    - Allows for enabling the benefits of 1 and 2.

A history item contains 2 things, its base search result (name, formatted address, coordinates), and an array of history. This "History array" has entries which also contain 2 things: a timestamp of when the search result was recorded, and a reason for _why_ the result was recorded. In other words, this history array records every time that a particular search result was saved in history, as well as the reason why for each time.

There are 3 reasons at the moment for recording search history:

1. The user hosted an event in the location.
2. The user attended an event in the location.
3. The user selected the location from the search screen.

Tracking these reasons allow for giving more context to results stored in the history. Say the user recently hosted an event at some random soccer field in say San Francisco. A few days later to search for a location for their next event. Since we've tracked a `"hosted-event"` reason for the soccer event in San Francisco, whenever the search screen displays that soccer field, we can annotate the result with something like, ***"You hosted an event here recently"***. Similar benefits exist for the other 2 reasons.

A couple of things that may need to be considered in the future:

1. Currently, there's no way to purge inactive search results.
2. Since we reference history items by location coordinates (since the location is the only guaranteed non-null value to be returned from Amplify Geo), this may prove problematic if a particular location has multiple coordinates. In that case it may be better to reference history items by address, and just not store locations with undefined addresses.

As per usual, a few generic utility functions were also able to be abstracted out of this functionality, mostly some Array stuff.

(PS: This also upgrades jest to v29, since there is some weird bug with `useFakeTimers` and async tests on previous versions)

### DIFF
--- a/lib/Array.ts
+++ b/lib/Array.ts
@@ -1,0 +1,8 @@
+export namespace ArrayUtils {
+  /**
+   * A typesafe way for filtering all null/undefined values from an array.
+   */
+  export const takeNonNulls = <T>(arr: (T | null | undefined)[]) => {
+    return arr.reduce((acc, curr) => (curr ? [...acc, curr] : acc), [] as T[])
+  }
+}
--- a/lib/Array.ts
+++ b/lib/Array.ts
@@ -1,8 +1,20 @@
+/**
+ * A helper type to ensure an array has at least 1 element.
+ */
+export type NonEmptyArray<T> = [T, ...T[]]
+
 export namespace ArrayUtils {
   /**
    * A typesafe way for filtering all null/undefined values from an array.
    */
   export const takeNonNulls = <T>(arr: (T | null | undefined)[]) => {
     return arr.reduce((acc, curr) => (curr ? [...acc, curr] : acc), [] as T[])
+  }
+
+  /**
+   * Returns the last element of a non-empty array.
+   */
+  export const lastElementNonEmpty = <T>(arr: NonEmptyArray<T>) => {
+    return arr[arr.length - 1]
   }
 }
--- a/lib/Array.ts
+++ b/lib/Array.ts
@@ -5,9 +5,9 @@ export type NonEmptyArray<T> = [T, ...T[]]
 
 export namespace ArrayUtils {
   /**
-   * A typesafe way for filtering all null/undefined values from an array.
+   * A typesafe way for removing all null/undefined values from an array.
    */
-  export const filterOptionals = <T>(arr: (T | null | undefined)[]) => {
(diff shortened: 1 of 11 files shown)


lib/Array.ts
```diff
@@ -0,0 +1,20 @@
...
+ * A helper type to ensure an array has at least 1 element.
+ */
+export type NonEmptyArray<T> = [T, ...T[]]
+
+export namespace ArrayUtils {
+  /**
+   * A typesafe way for removing all null/undefined values from an array.
+   */
+  export const removeOptionals = <T>(arr: (T | null | undefined)[]) => {
+    return arr.reduce((acc, curr) => (curr ? [...acc, curr] : acc), [] as T[])
```
seanim0920, line 11: We could also use .filter() to remove the null/undefined values

lib/location/search/History.ts
```diff
@@ -0,0 +1,197 @@
...
+ * An interface for accessing the user's location search history.
+ */
+export interface LocationSearchHistory {
+  /**
+   * Loads all history items sorted by the most recently saved item timestamp
+   * in descending order.
+   *
+   * @param options see {@link LocationSearchHistoryLoadOptions}
+   */
+  loadItems: (
```
seanim0920, line 100: Very easy to use API 👍

seanim0920 (approved):

lib/Array.ts
mhayes853: `filter` doesn't actually change the type however, the return would still be `(T | undefined | null)[]`, which is why a `reduce` was needed.

ZoroarkZark (approved):

nicoletteL (approved):

PR #72, opened 2023-03-10
Cleanup: Represent Null values on Placemark as undefined

mhayes853:
Since js has both `null` and `undefined`, and react tends to prefer `undefined` for optional prop values, I decided to covert the `Placemark` type to hold a bunch of possible undefined values instead of a bunch of `string | null` types. This removed the ridiculous `placemark.someField ?? undefined` dance that was performed several times.

Additionally, I also added a generic utility to convert null values on any object to undefined, which should make life easier when working with react props. It's called `nullsToUndefined`, and it will mostly be useful in turning 3rd party library types that use null into ones that use undefined.

### DIFF
--- a/lib/NullsToUndefined.ts
+++ b/lib/NullsToUndefined.ts
@@ -1,0 +1,37 @@
+/**
+ * A type which ensures that all `null` values become `undefined`.
+ *
+ * See: https://stackoverflow.com/questions/50374869/generic-way-to-convert-all-instances-of-null-to-undefined-in-typescript
+ */
+export type NullsAsUndefined<T> = T extends null
+  ? undefined
+  : T extends Date
+  ? T
+  : {
+      [K in keyof T]: T[K] extends (infer U)[]
+        ? NullsAsUndefined<U>[]
+        : NullsAsUndefined<T[K]>
+    }
+
+/**
+ * Converts null values of an object to undefined.
+ *
+ * @param value the object
+ */
+export const nullsToUndefined = <T>(value: T) => {
+  if (value === null) return undefined as NullsAsUndefined<T>
+  if (typeof value !== "object") return value as NullsAsUndefined<T>
+
+  // NB: For the bottom 2 cases we must extract out a local variable instead
+  // of returning directly to make the compiler happy.
+
+  if (value instanceof Array) {
+    const array = value.map(nullsToUndefined) as NullsAsUndefined<T>
+    return array
+  }
+
+  const object = Object.fromEntries(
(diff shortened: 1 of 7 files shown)


nicoletteL (approved):

nicoletteL:
Thanks, this will definitely be useful.

lib/NullsToUndefined.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * A type which ensures that all `null` values become `undefined`.
+ *
+ * See: https://stackoverflow.com/questions/50374869/generic-way-to-convert-all-instances-of-null-to-undefined-in-typescript
+ */
+export type NullsAsUndefined<T> = T extends null
```
seanim0920, line 6: There may be cases where we want to explicitly mark something as "null". Can't think of any atm though, just something to look out for in the future

lib/location/Geocoding.ts
```diff
@@ -32,10 +33,11 @@ export class ExpoGeocoding implements Geocoding {
   }
 
   async reverseGeocode (location: Location) {
-    const results = await ExpoLocation.reverseGeocodeAsync({
-      ...location
-    })
-    return results as Placemark[]
+    return nullsToUndefined(
```
seanim0920, line 36: Not sure if we want/need to use this function gratuitously

tests/location/PlacemarkFormatting.test.ts
```diff
@@ -10,13 +10,13 @@ describe("Placemark Formatting tests", () => {
 
   it("omits street number when missing street name", () => {
     expect(
-      placemarkToFormattedAddress({ ...baseTestPlacemark, street: null })
+      placemarkToFormattedAddress({ ...baseTestPlacemark, street: undefined })
```
seanim0920, line 13: Makes sense

seanim0920 (approved):

lib/NullsToUndefined.ts
mhayes853: Maybe, but let's consider that most other languages don't have a divide between null and undefined, but rather just a unified null value. Would prefer to lean towards keeping everything as undefined if possible since undefined has a nicer syntax than dealing with null.

lib/location/Geocoding.ts
mhayes853: I probably should add a note about this in the docs for it, but generally I agree. Its only real use-case imo is for converting types from 3rd party libraries that use null, however it's still a nice to have I think.

ZoroarkZark (approved):

PR #73, opened 2023-03-11
Event UI for Menu Dropdown and Delete Alert

nicoletteL:
This PR includes some slight changes to the event UI (lighter shadow using the event color, consistent spacing) as well as dropdown menu functionality to the more button. 

If a user is the host of the event, pressing the more button will bring up options to either delete or edit an event. Right now pressing delete only shows an alert asking if the user is sure they want to delete the event (no deletion functionality). Pressing edit doesn't do anything other than send a console message.

If the user isn't the host, pressing the more button brings up an option to report the post. Right now reporting does the same thing as deletion (causes an alert), but I think in the future we'll want it to be a modal with reasons a user can select for reporting (or a text entry describing why).

The menu is implemented using react-native-popup-menu. It has open/close functionality already built in, however, due to the fact that testID doesn't seem work on Menu Option currently I had to pass the onPress functions to the children so the functionality could be tested. The only thing this caused was having to have state to open/close the menu manually.

Also, I was able to add a greyed-out backdrop to the screen when the menu is selected so the user knew that the screen wasn't interactable when the menu was open.


components/EventsList.tsx
```diff
@@ -24,11 +25,15 @@ const EventsList = () => {
   }, [])
 
   // variables
-  const snapPoints = ["4%", "65%", "100%"]
+  const snapPoints = ["3%", "65%", "100%"]
```
mhayes853, line 28: We'll probably want the min height to be more like 10% (so that the user can see the "Nearby Activities" text) and the last snap point to be more like 75% (100% makes the sheet practically unusable on iOS)

components/eventCard/EventItem.tsx
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  // Will probably need to use a modal instead to have user select/type in reason for report
+  const alertReport = () => {
+    dismissMenu()
+    Alert.alert("Report this event?", undefined, [
+      { text: "Report", style: "destructive", onPress: dismiss },
+      { text: "Cancel", style: "cancel" }
+    ])
+  }
```
mhayes853, line 51: Is it possible that we can get the menu stuff into a reusable component? We'll need it for the detail view.

tests/events/EventCard/eventUI.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+        />
+      </MenuProvider>
+    )
+    fireEvent.press(screen.getByTestId(moreButtonLabel))
+    fireEvent.press(screen.getByText("Delete"))
+
+    expect(alertPresentationSpy).toHaveBeenCalled()
+  })
+
+  //* * Right now event isn't removed */
```
mhayes853, line 90: Suggestion: Since this deletion action will probably only ever have 1 implementation, you may as well fire off a react query mutation that calls some kind of deletion method on the `Events` interface.

You can then assert on whether or not the deletion happened by providing a test `Events` interface using `SetDependencyValue`.

tests/events/EventCard/eventUI.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+        />
+      </MenuProvider>
+    )
+    fireEvent.press(screen.getByTestId(moreButtonLabel))
+    fireEvent.press(screen.getByText("Report"))
+
+    dismissConfirmReport()
+    expect(screen.queryByText("Report this event?")).toBeNull()
+  })
+})
```
mhayes853, line 147: Tests are thorough ✅

I'll approve once we separate out the menu stuff into a reusable component and move these tests as such.

components/eventCard/MenuDropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+  MenuTrigger
+} from "react-native-popup-menu"
+import { TouchableOpacity } from "react-native-gesture-handler"
+import { MaterialIcon } from "@components/common/Icons"
+
+interface Props {
+  isEventHost: boolean
+}
+
+const MenuDropdown = ({ isEventHost }: Props) => {
```
mhayes853, line 16: I think this could use a more specific name related to events, but thanks for this.

tests/events/EventCard/menuDropdown.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+    render(
+      <MenuProvider>
+        <MenuDropdown isEventHost={true} />
+      </MenuProvider>
+    )
+    fireEvent.press(screen.getByTestId(moreButtonLabel))
+    fireEvent.press(screen.getByText("Delete"))
+
+    dismissConfirmDelete()
+    expect(screen.queryByText("Delete this event?")).toBeNull()
```
mhayes853, line 69: One suggestion, since you've extracted `dismissConfirmDelete` which taps the delete button, you may also want to do this for the `fireEvent` stuff as well. In other words, I think the test becomes more self-explanatory when it looks like this (using a hypothetical `Events` interface):

```ts
render(...)
attemptDeletion()
await confirmDeletion()
// Assert that the event was actually deleted in some way
```

rather than

```ts
render(...)
fireEvent.press(screen.getByTestId(moreButtonLabel))
fireEvent.press(screen.getByText("Delete"))
dismissConfirmDelete()
// Assert that the event was actually deleted in some way
```

The former just presents the test as more of a set of requirements rather than interacting directly with implementation details.

tests/events/EventCard/menuDropdown.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+})
+
+// Labels
+const moreButtonLabel = "more options"
+
+// Alert helpers
+const { alertPresentationSpy, tapAlertButton } = captureAlerts()
+
+const dismissConfirmDelete = () => {
+  tapAlertButton("Delete")
```
mhayes853, line 106: Btw just a heads up that this `tapAlertButton` is actually an `async` function, so you should await it in this otherwise it may cause problems for future tests

mhayes853 (approved):

tests/events/EventCard/menuDropdown.test.tsx
nicoletteL: Thanks for the suggestion! It does look a lot cleaner and more intuitive to do it that way.

ZoroarkZark (dismissed):

mhayes853 (approved):

ZoroarkZark (approved):

goldenwarrior2 (approved):

PR #74, opened 2023-03-11
event schema complete

goldenwarrior2:


seanim0920:
Also include that the graphql v2 migration is here

PR #76, opened 2023-03-12
Location Text Searching

mhayes853:
This PR simply includes a function which calls out to Amplify Geo under the hood to perform text based location searches. There is some significant handling that needed to be performed to parse the response from amplify geo which is handled in a separate utility function that even has its own set of tests.


ZoroarkZark (approved):

seanim0920:
A lot of these diffs are due to the new backend updates, not sure if it will conflict with Tristans PR

seanim0920:
We should their PR merged first so this one will be easier to parse, I think they'll still need to rebase off development

seanim0920:
Alternatively you could discard all of the backend changes besides the ones relating to amplify geo

mhayes853:
Going to close this and reopen with some changes to the `Placemark` type, since the code here is essentially a better version of a normal geocoder.

PR #78, opened 2023-03-13
Fix babel storybook issue

seanim0920:
Imports in storybook should work as expected now by using the same babel config as the root project.

Remaining issues:
* VS still doesn't recognize the @path shortcuts in storybook, but they should build properly in expo
* Storybook adds a margin on top of the stacknavigator for some reason. Adding "headerStatusBarHeight" to 0 fixes it but it would need to go in the stack navigator directly, which I didn't include here.

### DIFF
--- a/.storybook/.ondevice/Storybook.tsx
+++ b/.storybook/.ondevice/Storybook.tsx
@@ -1,5 +1,7 @@
-import { getStorybookUI } from "@storybook/react-native";
-import "./storybook.requires";
+import { getStorybookUI } from "@storybook/react-native"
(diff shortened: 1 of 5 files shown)


ZoroarkZark (approved):

nicoletteL (approved):

mhayes853 (approved):
I'm still receiving import issues but only on iOS now. Also the bottom sheet library seems to crash the entire expo go app for me only through storybook, again specifically on iOS.

However, we can probably just deal with those issues later, since I don't want anyone else to get blocked atm.

PR #84, opened 2023-03-15
Backend event notifications

nairoldi:


seanim0920:
Converted to draft so we can use the new backend systems

ZoroarkZark:
New backend

PR #85, opened 2023-03-19
Refactor: Event Type

mhayes853:
This PR makes the `Event` type much more flexible, by improving its separation of concerns such that the type contains only the bare necessities (eg. Host and basic info). The `Event` type does not contain any information on the current user's perspective of the event, but rather there's a type called `CurrentUserEvent` which provides 2 additional properties. One for whether or not the user is a host, and another for their distance from the event (in miles for now as we can support a more generic Metric type once we move out of the US). 

Additionally, since a user can only either be a "host", "regular attendee", or a "non-participant", there's a union type called `EventCurrentUserAttendeeStatus` which fills the first additional property on `CurrentUserEvent`. This type also has a couple of convieneince functions to check whether or not the user is either a host or attending an event so we don't have a lot of `event.userAttendeeStatus !== "hosting"` nonsense.

I also renamed the `Location` type to `LocationCoordinate2D` since it only represents longitude and latitude.

Finally, I've removed the `Events` interface for the time being, since nothing was actually using it to any decent degree, this should give us some more time to decide on how events are fetched from an interface (I mean interface in the general sense, not referring to a TS interface) perspective.

PS: I also added some mock events and event attendees... Since this isn't a book club, I decided that Harrison's infamous creation of Molly Member has died, and Anna Admin changed her name to Anna Attendee bc she murdered Molly and is hiding from law enforcement. So Haley Host is now taking the role of Molly, as well as these 3 Blob, Blob Jr., and Blob Sr. things.


components/eventCard/EventCard.tsx
nicoletteL, line 1: :+1: Thanks for renaming and organizing the events.

lib/events/Event.ts
nicoletteL, line 1: Thanks for breaking up the event type since it'll make the type more flexible. The functions to check the user's status is also very helpful.

nicoletteL (approved):

ZoroarkZark (approved):

PR #87, opened 2023-03-19
Update: Make a test fail when it uses the default value from a dependency key

mhayes853:
This PR requires that `SetDependencyValue` or `UpdateDependencyValue` is used for all dependency keys in a given test context. If a dependency key attempts to use its default value inside a test, the `DependencyValues` class will throw an error (which is highly likely to cause a test failure).

There are 2 reasons for why this is a good idea:
1. It prevents tests from accidentally making AWS calls, tracking analytics, or doing other costly operations.
2. It ensures that when you write tests, you don't miss covering important side effects. By explicitly requiring that you give a preset value for a dependency key, you will naturally think about how the component under test uses the dependency.

### DIFF
--- a/tests/dependencies/helpers.ts
+++ b/tests/dependencies/helpers.ts
@@ -1,0 +1,11 @@
+/**
+ * Only for use in dependencies tests which need to test default value behavior.
+ */
+export const _negateDependencyKeyDefaultValueTestContext = () => {
+  let jestId: string | undefined
+  beforeAll(() => {
+    jestId = process.env.JEST_WORKER_ID
+    process.env.JEST_WORKER_ID = undefined
+  })
+  afterAll(() => (process.env.JEST_WORKER_ID = jestId))
+}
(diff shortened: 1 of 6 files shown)


tests/dependencies/helpers.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Only for use in dependencies tests which need to test default value behavior.
+ */
+export const _negateDependencyKeyDefaultValueTestContext = () => {
```
seanim0920, line 4: Would be good to have a rule of some sort to prevent people from using this gratuitously

seanim0920 (approved):

tests/dependencies/helpers.ts
mhayes853: Sure, though I think the underscore should be enough to scare people away.

nicoletteL (approved):

ZoroarkZark (approved):

PR #88, opened 2023-03-27
useUserLocation hook

mhayes853:
This PR just includes a simple way to track the current user's location with a hook. Note that the location can be undefined if the the user has denied permissions, has bad network, etc.

### DIFF
--- a/lib/dependencies/DependencyValues.ts
+++ b/lib/dependencies/DependencyValues.ts
@@ -72,7 +72,10 @@ export class DependencyValues {
     const cachedValue = this.cachedValues.get(key.identifier)
     if (cachedValue) return cachedValue as T
 
-    if (process.env.JEST_WORKER_ID !== "undefined") {
+    if (
+      process.env.JEST_WORKER_ID !== "undefined" &&
+      process.env.JEST_WORKER_ID !== undefined
+    ) {
       throw new Error(`
       Attempted to access the default value of a dependency key in a test context.
 
--- a/lib/location/UserLocation.ts
+++ b/lib/location/UserLocation.ts
@@ -1,6 +1,14 @@
 import { TrackedLocation } from "./Location"
-import ExpoLocation from "expo-location"
-import { createDependencyKey } from "../dependencies"
+import {
+  watchPositionAsync,
+  requestForegroundPermissionsAsync,
+  getLastKnownPositionAsync,
+  getCurrentPositionAsync,
+  LocationObject,
+  LocationAccuracy
+} from "expo-location"
+import { createDependencyKey, useDependencyValue } from "../dependencies"
+import { useEffect, useRef, useState } from "react"
 
 export type StopUserLocationTracking = () => void
 
@@ -95,23 +103,23 @@ export interface UserLocation {
  */
 export class ExpoUserLocation implements UserLocation {
   async requestForegroundPermission () {
-    return (await ExpoLocation.requestForegroundPermissionsAsync()).granted
+    return (await requestForegroundPermissionsAsync()).granted
   }
 
   async lastKnownLocation () {
-    const location = await ExpoLocation.getLastKnownPositionAsync()
+    const location = await getLastKnownPositionAsync()
     return location ? toTrackedLocation(location) : null
   }
 
   async currentLocation () {
-    return toTrackedLocation(await ExpoLocation.getCurrentPositionAsync())
+    return toTrackedLocation(await getCurrentPositionAsync())
   }
 
   async track (
     callback: (location: TrackedLocation) => void,
     options: UserLocationTrackingOptions = defaultTrackingOptions
   ) {
-    const subscription = await ExpoLocation.watchPositionAsync(
+    const subscription = await watchPositionAsync(
       {
         accuracy: toExpoAccurracy(options.accuracy),
         distanceInterval: options.minUpdateMetersDistance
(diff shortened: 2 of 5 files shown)


ZoroarkZark (approved):

nicoletteL (dismissed):

mhayes853:
**Small notice**
When a user who previously denied location permissions re-enables them in settings, expo doesn't pick up said update to the permissions, and as a result still gives back error updates. The fix is to re-request the location permissions. Afaik, this only happens on android (galaxy s10e) btw. I've documented this in the hook.

lib/location/UserLocation.ts
```diff
@@ -1,154 +1,99 @@
...
+  coordinates: {
+    latitude: location.coords.latitude,
+    longitude: location.coords.longitude
+  },
+  trackingDate: new Date(location.timestamp / 1000)
+})
 
 /**
- * `DependencyKey` for `UserLocation` operations.
+ * Dependency Keys relating to operations around the user's location.
```
seanim0920, line 90: Makes sense

lib/location/UserLocation.ts
```diff
@@ -1,154 +1,99 @@
...
-    coordinate: locationResponse.coords,
-    trackingDate: new Date(locationResponse.timestamp)
-  } as TrackedLocation
-}
+const expoLocationToTrackedLocation = (location: LocationObject) => ({
+  coordinates: {
+    latitude: location.coords.latitude,
+    longitude: location.coords.longitude
+  },
+  trackingDate: new Date(location.timestamp / 1000)
```
seanim0920, line 86: Is this necessary for Expo to work or is it for us to use

lib/location/UserLocation.ts
```diff
@@ -1,154 +1,99 @@
...
-      },
-      (location) => callback(toTrackedLocation(location))
-    )
-    return subscription.remove as StopUserLocationTracking
-  }
-}
-
-const defaultTrackingOptions: UserLocationTrackingOptions = {
-  accuracy: UserLocationTrackingAccurracy.Medium
+export const expoTrackUserLocation = (
```
seanim0920, line 54: It looks like this isn't intended to be used outside of this file and only called through the dependency system. Is there a way we can prevent accidental direct import/misuse of this function in the future? We could remove "export" but we'd need that for testing.

lib/dependencies/DependencyValues.ts
```diff
@@ -72,7 +72,10 @@ export class DependencyValues {
     const cachedValue = this.cachedValues.get(key.identifier)
     if (cachedValue) return cachedValue as T
 
-    if (process.env.JEST_WORKER_ID !== "undefined") {
+    if (
+      process.env.JEST_WORKER_ID !== "undefined" &&
```
seanim0920, line 76: I'm surprised we need both the string and value of undefined checked here!

tests/location/UserLocationHooks.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+          value={trackUserLocation}
+        >
+          {children}
+        </SetDependencyValue>
+      )
+      return renderHook(useTrackUserLocation, { wrapper })
+    }
+
+    test("processes simple location updates", () => {
+      let sendLocationUpdate: (update: UserLocationTrackingUpdate) => void
```
seanim0920, line 34: Are there any edge cases we'll need to test in this module, or is this just basic implementation

tests/location/UserLocation.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  expoTrackUserLocation,
+  LocationCoordinatesMocks,
+  UserLocationTrackingAccurracy
+} from "@lib/location"
+
+const testAccuracy = "precise"
+
+describe("UserLocation tests", () => {
+  describe("ExpoUserLocationTracking tests", () => {
+    test("translates accurracy to proper expo accurracy when making tracking request", () => {
```
seanim0920, line 12: It would be better not to need a unit test for mapping, js gives us plenty of options to have a mapping between strings (enums, objects)

tests/location/UserLocation.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      // expo's tracker returns a promise). This promise ensures that we give
+      // a chance for the tracker to resolve before asserting.
+      const trackPromise = new Promise<void>((resolve) => {
+        track.mockImplementation(() => {
+          resolve()
+          return Promise.resolve({ remove: unsubFromExpo })
+        })
+      })
+
+      const unsub = expoTrackUserLocation(testAccuracy, jest.fn(), track)
```
seanim0920, line 72: We can also leave this out since there aren't really edge cases here, it's just checking react native's useEffect cleanup. Also people will likely not use expoTrackUserLocation directly so manual unsubbing isn't something to worry about.

seanim0920 (dismissed):

lib/location/UserLocation.ts
mhayes853: Unfortunately ts doesn't have precise access control... Otherwise yeah I would agree.

tests/location/UserLocation.test.ts
mhayes853: You're right about the case where inside a `useEffect` it would be useless to test (since we get the cleanup logic for free). However, in this case the test is testing for what the unsubscribe function _actually_ does since this test isn't inside a react component. In this case, it simply means, once we have subscribed to the tracker (which may not be the case when `unsub` is called, and thus we have to wait), do we actually unsubscribe.

mhayes853:
I didn't know this but apparently the `waitFor` test helper that comes with the react native testing library can also be used outside of react component tests. That makes things a bit cleaner.

seanim0920 (approved):

ZoroarkZark (approved):

PR #90, opened 2023-04-08
Reworked UI for Event Card

nicoletteL:
This PR implements the UI designs shown during our meetings. This includes the removal of the attendees, distance, and drop shadow on the card. I've also changed the font to use Open Sans.


components/EventsList.tsx
```diff
@@ -69,10 +81,10 @@ const styles = StyleSheet.create({
     paddingBottom: 8
   },
   activitiesText: {
-    fontSize: 20,
+    fontSize: 18,
     marginLeft: 16,
     textAlignVertical: "top",
-    fontWeight: "bold"
+    fontFamily: "Open-Sans-Bold"
```
seanim0920, line 87: @mhayes853  planned to make a separate <Text> component with this styling built-in, sync with him

lib/location/Placemark.ts
```diff
@@ -44,4 +44,13 @@ export const placemarkToFormattedAddress = (placemark: Placemark) => {
   return formattedAddress === US_COUNTRY_CODE ? undefined : formattedAddress
 }
 
+/**
+ * Formats a `Placemark` instance into a short address (name, city, region/state)
+ *
+ * At the moment, this only outputs US style addresses.
+ */
+export const placemarkToAbbreviatedAddress = (placemark: Placemark) => {
+  return `${placemark.name}, ${placemark.city}, ${placemark.region}`
```
seanim0920, line 53: In case any of these are null we should have fallbacks @mhayes853

lib/date/DateRange.ts
```diff
@@ -114,7 +110,20 @@ const formatDate = (basis: Dayjs, date: Dayjs) => {
     return date.format("ddd")
   }
   const yearFormat = !date.isSame(basis, "year") ? " YYYY" : ""
-  return date.format(`MMM D${yearFormat}`)
+  return date.format(`ddd, MMM D${yearFormat}`)
+}
+
+const formatFromBasis = (basis: Dayjs, date: Dayjs) => {
```
seanim0920, line 116: I'm not sure what "basis" means, is there a better name

components/eventCard/EventCard.tsx
```diff
@@ -108,62 +102,44 @@ const styles = StyleSheet.create({
...
   image: {
-    width: 40,
-    height: 40,
+    width: 32,
+    height: 32,
     borderRadius: 24
   },
   container: {
-    backgroundColor: "white",
+    backgroundColor: "#f5f5f5",
```
seanim0920, line 142: We should have this in a separate file so it's easy to reuse in other components

components/eventCard/EventCard.tsx
```diff
@@ -1,104 +1,98 @@
...
-            style={[styles.image, styles.iconMargin]}
-            source={require("../../assets/icon.png")}
-            accessibilityLabel="profile picture"
-          />
+    <View style={styles.container}>
+      <View style={[styles.topRow, styles.flexRow]}>
+        <Image
+          style={[styles.image, styles.iconMargin]}
+          source={require("../../assets/icon.png")}
+          accessibilityLabel="profile picture"
```
seanim0920, line 32: Ty for accessibility label. We'll need a separate component for the profile image at least since I'm sure we'll reuse that a lot in different screens.

components/eventCard/EventCard.tsx
```diff
@@ -1,104 +1,98 @@
...
-  const lightEventColor = tinycolor(event.color).lighten(25).toString()
+  const hexAlpha = "4D"
+  const lightEventColor = event.color + hexAlpha
+
+  const [fontsLoaded] = useFonts({
+    "Open-Sans": require("../../assets/fonts/OpenSans-Regular.ttf"),
+    "Open-Sans-Bold": require("../../assets/fonts/OpenSans-Bold.ttf")
+  })
+
+  if (!fontsLoaded) {
```
seanim0920, line 22: It would be nice if we didn't have to have this check everywhere although not sure how to handle that atm

seanim0920 (dismissed):

lib/location/Placemark.ts
mhayes853: Maybe something like:
- No Name -> Just render city and region
- No Region or City -> Render "Unknown Location"

lib/date/DateRange.ts
```diff
@@ -83,26 +83,22 @@ export class FixedDateRange {
...
-  if (date.isBetween(current, current.add(7, "days"), "days")) {
-    return date.format("ddd") + ` ${formattedTime}`
+  formattedStartTime () {
+    const start = dayjs(this.startDate)
+    return start.format("h:mm A")
+  }
+
+  formattedStartDate () {
+    const start = dayjs(this.startDate)
+    return formatDate(now(), start)
```
mhayes853, line 100: We should probably just have general functions to format a single date and just call `formatDate(dateRange.startDate)` when we need to in specific areas. I think we have way too many of these formatter functions on the `FixedDateRange` class when it doesn't really make sense in a lot of cases.

components/EventsList.tsx
```diff
@@ -5,23 +5,35 @@ import {
...
+    EventMocks.PickupBasketball
+  ]
+  const MARGIN_HORIZONTAL = 24
   const MARGIN_VERTICAL = 16
 
   // hooks
   const sheetRef = useRef<BottomSheetModal>(null)
+  const [fontsLoaded] = useFonts({
+    "Open-Sans-Bold": require("../assets/fonts/OpenSans-Bold.ttf")
+  })
```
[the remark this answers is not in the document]
nicoletteL: Yeah, I started thinking about the fonts earlier and realized that loading it in the components wasn't good. I agree that putting it in its own hook and calling it at the root is the best option.

I didn't realize that expo fonts have first class support for google fonts. That's nice.

suryatiru (dismissed):

ZoroarkZark (approved):

app.json
```diff
@@ -34,6 +34,7 @@
     "splash": {
       "image": "./assets/splash.png",
       "backgroundColor": "#26C6A2"
-    }
+    },
+    "assets": ["./assets/fonts/"]
```
mhayes853, line 38: You can remove this line since we already have installed the fonts through google fonts.

lib/location/Placemark.ts
```diff
@@ -44,4 +44,18 @@ export const placemarkToFormattedAddress = (placemark: Placemark) => {
...
+/**
+ * Formats a `Placemark` instance into a short address (name, city, region/state)
+ *
+ * At the moment, this only outputs US style addresses.
+ */
+export const placemarkToAbbreviatedAddress = (placemark: Placemark) => {
+  if (placemark.city === undefined || placemark.region === undefined) {
+    return "Unknown Location"
+  } else if (placemark.name === undefined) {
+    return `${placemark.city}, ${placemark.region}`
```
mhayes853, line 56: You can use `!placemark.xxx` to check if something is undefined in a more concise way. Additionally, I think this function is only used for the event card so you can definitely just move it in the same file.

mhayes853 (approved):
Some next steps:
1. Utilize the new text components (it will fix some of the incorrect font sizes).
2. The colored icons don't have enough padding.
3. There should be some way to respond to when the user's profile is tapped (say via a callback prop)

mhayes853 (approved):
Some next steps:
1. Utilize the new text components (it will fix some of the incorrect font sizes).
2. The colored icons don't have enough padding.
3. There should be some way to respond to when the user's profile is tapped (say via a callback prop)

PR #91, opened 2023-04-08
UI for the Event Tab Bar

nicoletteL:
This PR includes a component for the tab bar at the bottom of the Activities Screen. I tried to implement glassmorphism/blur under the tab bar, but it doesn't seem like Android supports this very well. There is a library called react-native-community/blur that supports it for cross-platform, but I couldn't get it to work (I don't think it works with expo as the solution seems to be to edit the Android gradle files).

Expo has its own library which I'm using, but right now for Android it just uses a semi-transparent view. It looks like they've just added support on Android though so I think blurring will be supported in the next build.

I haven't tested on IOS so I don't know how well the blurring works, so feel free to modify it.


mhayes853:
If it comes down to it, I could always just write a native view wrapper, as iOS has builtin glassmorphic views, and it's not to hard to replicate on android (with say jetpack compose).

Though the only thing special about glassmorphism is the background blur, so that's really the problem we need to solve here.

components/tabBarComponents/EventTabBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { BlurView } from "expo-blur"
+
+export const EventTabBar = () => {
+  const darkIconColor = "#1c1c1c"
+  const iconSize = 24
+
+  return (
+    <BlurView intensity={100} tint="light" style={styles.blur}>
+      <View style={styles.container}>
+        <TouchableOpacity style={styles.icon}>
```
seanim0920, line 13: We'll need some hitSlop for each button to make them easier to tap. Cover in a follow up PR

components/tabBarComponents/EventTabBar.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react"
+import { StyleSheet, TouchableOpacity, View } from "react-native"
+import { Ionicons } from "@expo/vector-icons"
+import { BlurView } from "expo-blur"
+
+export const EventTabBar = () => {
```
seanim0920, line 6: Rename to "NavigationTabBar" or something similar, since this will be used for navigation across the whole app

seanim0920 (approved):

components/tabBarComponents/EventTabBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <Ionicons name="location" size={iconSize} color={darkIconColor} />
+        </TouchableOpacity>
+        <TouchableOpacity style={styles.icon}>
+          <Ionicons name="chatbox" size={iconSize} color={"grey"} />
+        </TouchableOpacity>
+        <TouchableOpacity style={styles.icon}>
+          <Ionicons name="add-circle" size={60} color={darkIconColor} />
+        </TouchableOpacity>
+        <TouchableOpacity style={styles.icon}>
+          <Ionicons name="notifications" size={iconSize} color={"grey"} />
```
mhayes853, line 23: Use black with an opacity of say 0.35 instead of a gray color. It's easier to work with, and makes supporting things like dark mode easier in the future.

package.json
```diff
@@ -51,6 +51,7 @@
     "dayjs": "^1.11.7",
     "expo": "^46.0.0",
     "expo-av": "~12.0.4",
+    "expo-blur": "^12.2.2",
```
mhayes853, line 54: For now, a simple white background will be fine if react native hates glassmorphism.

mhayes853 (approved):
The next step will be making this into something that works with react navigation.
Look into [this](https://reactnavigation.org/docs/bottom-tab-navigator#tabbar).

PR #94, opened 2023-04-10
Map Split + EventDetails Map

ZoroarkZark:
For this PR, I edited the Map component to be EventsMap and DetailsMap, both of which are custom components meant to be used for the ActivitiesScreen and EventDetails screen. In addition, I implemented the DetailsMap into the EventDetails component. 

At the moment, I'm not sure if we should be editing the EventDetails to be a screen instead of a component, so I would love some feedback on that aspect.


components/eventDetails/MapSnippet.tsx
```diff
@@ -0,0 +1,100 @@
...
+      {
+        <>
+          <Marker
+            key={marker.key}
+            coordinate={{
+              latitude: marker.location.latitude,
+              longitude: marker.location.longitude
+            }}
+          >
+            {renderMarker && renderMarker(marker)}
```
ZoroarkZark, line 92: Try and allow for a profile image to be displayed instead, and that be the main thing to be shown instead; no need for a renderMarker method

components/eventDetails/MapSnippet.tsx
```diff
@@ -0,0 +1,100 @@
...
+    latitudeDelta: number
+    longitudeDelta: number
+  }
+
+  /**
+   * A single map marker to display.
+   */
+  marker: T
+
+  /**
```
ZoroarkZark, line 32: Remove these from the mapSnippet, redundant since settings are static

components/eventDetails/MapSnippet.tsx
```diff
@@ -0,0 +1,100 @@
...
+        },
+        {
+          featureType: "transit",
+          stylers: [{ visibility: "off" }]
+        }
+      ]}
+    >
+      {
+        <>
+          <Marker
```
ZoroarkZark, line 85: Read in the event, and use the information from that event directly instead of creating a marker object as a prop

components/eventDetails/MapSnippet.tsx
seanim0920: Yes, we can make this component as small as possible, no need to go generic here

components/eventDetails/EventDetails.tsx
```diff
@@ -124,6 +125,16 @@ const EventDetails = ({ event }: EventDetailsProps) => {
             </View>
           )}
       </View>
+      <MapSnippet
+        style={{ width: "100%", height: "30%" }}
```
seanim0920, line 129: I wonder if we want the height to be a percentage or a fixed number. If possible, find someone with a different screen size to test it and see if it looks good. @suryatiru If you're available, check it out on your device.

components/eventDetails/MapSnippet.tsx
```diff
@@ -0,0 +1,100 @@
...
+   */
+  canRotate?: boolean
+
+  /**
+   * Renders a replacement for a singular map marker that was passed in through `markers`.
+   */
+  renderMarker?: (marker: T) => React.ReactNode
+}
+
+export const MapSnippet = function Map<T extends EventMarker> ({
```
seanim0920, line 53: MapSnippet -> EventMapSnippet

components/EventsMap.tsx
```diff
@@ -155,13 +150,12 @@ export const Map = forwardRef(function ReffedMap<T extends MapMarker> (
             }}
             onPress={() => onMarkerSelected?.(marker)}
           >
-            {renderMarker(marker)}
+            {renderMarker && renderMarker(marker)}
```
seanim0920, line 153: We can also do {renderMarker?.(marker)}
It does the same thing

components/EventsMap.tsx
```diff
@@ -94,12 +88,11 @@ export type MapProps<T extends MapMarker> = {
 /**
  * A view for rendering a google map.
  */
-export const Map = forwardRef(function ReffedMap<T extends MapMarker> (
+export const EventsMap = forwardRef(function ReffedMap<T extends MapMarker> (
   {
     initialRegion,
     markers,
     renderMarker,
```
seanim0920, line 95: I wonder if we'll still need renderMarker too, or if we'll use the profile pic of each event similar to the map snippet

app.json
```diff
@@ -16,24 +16,25 @@
...
+      }
     },
     "web": {
       "favicon": "./assets/favicon.png"
     },
     "android": {
       "config": {
         "googleMaps": {
-          "apiKey": "<insert API key here>"
+          "apiKey": "insert key here"
```
seanim0920, line 31: Thanks

seanim0920 (approved):

components/eventDetails/EventDetails.tsx
```diff
@@ -124,6 +136,67 @@ const EventDetails = ({ event }: EventDetailsProps) => {
...
+        initialRegion={{
+          latitude: event.coordinates.latitude,
+          longitude: event.coordinates.longitude,
+          latitudeDelta: 0.1,
+          longitudeDelta: 0.1
+        }}
+        renderMarker={(item) => (
+          <>
+            {/* Solid white background for the image */}
+            <ImageBackground
```
ZoroarkZark, line 150: Don't go with white image, go with background color white

screens/ActivitiesScreen.tsx
```diff
@@ -1,18 +1,113 @@
...
+  }
   return (
     <>
-      <Map
+      <EventsMap
+        ref={appRef}
         style={{ width: "100%", height: "100%" }}
         initialRegion={state.initialRegion}
-        renderMarker={(item) => null}
+        renderMarker={(item) => (
```
ZoroarkZark, line 22: Make into component instead of declaring this big wall of text each time

screens/ActivitiesScreen.tsx
```diff
@@ -1,18 +1,113 @@
 import EventsList from "@components/EventsList"
-import Map from "@components/Map"
-import EventTabBar from "@components/tabBarComponents/EventTabBar"
-import React from "react"
+import EventsMap, { MapRefMethods } from "@components/EventsMap"
 import { state } from "@components/MapTestData"
+import EventTabBar from "@components/tabBarComponents/EventTabBar"
+import React, { useRef } from "react"
+import { ImageBackground, Text, TouchableOpacity, View } from "react-native"
```
ZoroarkZark, line 6: Use react-native-fast-image for loading up the profile image faster

seanim0920 (approved):

nicoletteL (approved):

PR #96, opened 2023-04-13
OpenSans Text Components

mhayes853:
#92 Added the open sans font and a hook to load the font. This PR now adds some preset text components since more design details have been fleshed out. Each text component takes in the standard `TextProps` so they can be used like the normal react native text component. Here's an image for reference:
IMG_2872

### DIFF
--- a/components/Text.tsx
+++ b/components/Text.tsx
@@ -1,0 +1,53 @@
+import React from "react"
+import { StyleSheet, Text, TextProps } from "react-native"
+
+/**
+ * A text component that represents the standard title of
+ * some UI element.
+ */
+export const Title = (props: TextProps) => (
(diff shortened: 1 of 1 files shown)


ZoroarkZark (approved):

components/Text.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const Caption = (props: TextProps) => (
+  <Text {...props} style={[props.style, styles.caption]} />
+)
+
+/**
+ * A text component with the same font size as {@link BodyText}
+ * that allows for more emphasis on a particular UI element
+ * group.
+ */
+export const Headline = (props: TextProps) => (
```
seanim0920, line 31: Add "bold caption" style

seanim0920 (approved):

PR #101, opened 2023-04-19
Button Components Shown in Design

nicoletteL:
This PR includes the button components shown in the designs and includes a dark variant and light variant. The title and onPress function are passed in through props so the components are reusable.
darkButton_1
lightButton_1

### DIFF
--- a/components/common/Buttons.tsx
+++ b/components/common/Buttons.tsx
@@ -1,38 +1,51 @@
+import { Headline } from "@components/Text"
+import { ButtonStyles } from "@lib/ButtonStyle"
 import React from "react"
-import {StyleSheet, TouchableOpacity, TouchableOpacityProps } from "react-native"
+import {
+  StyleSheet,
+  TouchableOpacity,
+  TouchableOpacityProps
+} from "react-native"
 
-const hexAlpha = "26" // 15% Opacity
-
-export namespace CustomButtonStyles {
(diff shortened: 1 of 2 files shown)


components/common/Buttons.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react"
+import { ButtonProps, Pressable, StyleSheet } from "react-native"
+import { Headline } from "@components/Text"
+
+const darkColor = "#26282A"
```
mhayes853, line 5: Should we maybe have a colors namespace (since this should be the text color as well, I guess I forgot to apply it in that PR)

components/common/Buttons.tsx
```diff
@@ -0,0 +1,51 @@
...
+const hexAlpha = "26" // 15% Opacity
+
+export const DarkButton = (props: ButtonProps) => {
```
mhayes853, line 8: I think `Button` and `OutlinedButton` are better names, but we may need a better name for the former since it clashes with react-native's buttons.

mhayes853 (approved):

nicoletteL:
I changed the component names to something more fitting (and I tried to avoid just calling it Button so it doesn't conflict with the native component), but feel free to suggest other names.

seanim0920 (approved):

seanim0920 (approved):

PR #102, opened 2023-04-19
Location Search UI

mhayes853:
This PR contains the UI for location searching.

Screenshot 2023-04-19 at 1 41 09 AM
Image 4-19-23 at 1 40 AM

The UI is split into 2 components, a search bar (very simple), and the actual picker itself (less simple). The 2 components utilize a jotai atom to share state between one another. This follows the designs since the search bar is naturally going to be the header of the respective screen that it's on.

There are also 2 dependency keys explicitly specified by the search picker, one to load the options, and another to save the option in the user's recent locations. The implementation of those is for a separate PR.

Additionally, I also added a query to get the current user's location once without the overhead of tracking. Tracking would cause the results to constantly update themselves (and burn more geocoding costs), which isn't ideal if say a user is searching locations inside a car going 60mph. One other thing is that we can probably share a cache between the tracker and this query, but we'll leave that for another time.

Lastly a bunch of old code was removed and replaced, with helpers as usual being added. There is still work to do on the data loading side...


mhayes853:
So I changed the `TestQueryClientProvider` to always require a query client that is created by `createTestQueryClient` so that we can't end up in this scenario. We'll also have to use a new `cleanupTestQueryClient` function inside of an `afterAll` to prevent these issues in the future.

mhayes853:
It seems that `Intl` is broken on android... Since we don't need multiple locales atm I just reimplemented the distance formatters with simple string interpolation, but it is kinda annoying that we can't use the official js one for now...

mhayes853:
It seems like setting the cache time to 0 when creating the test query client fixes the CI problem (hopefully for good...). With this, we should no longer have to pass in a custom test query client all the time (unless you need to test a mutation on it or something...).

seanim0920 (approved):

ZoroarkZark (approved):

PR #105, opened 2023-04-24
Current User Coordinates Expo Query

mhayes853:
This PR adds a wrapper around getting the user's current coordinates similar to the way we did it with tracking.

### DIFF
--- a/hooks/UserLocation.ts
+++ b/hooks/UserLocation.ts
@@ -3,8 +3,9 @@ import {
   UserLocationTrackingAccurracy,
   expoTrackUserLocation,
   TrackUserLocation,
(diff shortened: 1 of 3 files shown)


ZoroarkZark (approved):

hooks/UserLocation.ts
```diff
@@ -23,8 +24,9 @@ export namespace UserLocationDependencyKeys {
   /**
    * A dependency key to load the user's current coordinates a single time.
```
seanim0920, line 25: Maybe add a specific example of how it would be used

seanim0920 (approved):

PR #106, opened 2023-04-25
API Client

seanim0920:
+Added environment variables

* Currently saves auth info to globalThis
* Auth info will only save on sign in, and is put in asyncstorage for persistence


api-client/APIClient.test.ts
```diff
@@ -0,0 +1,98 @@
+import { z } from "zod"
+import { createApiClient } from "./api-client"
+import { GenericApiClient } from "./api-endpoint-types"
+
+class Response {
```
seanim0920, line 5: React Native should have it defined but jest didnt recognize it

api-client/APIClient.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  })
+
+  it("calls the API with the correct method, endpoint, headers, and path params", async () => {
+    const apiClient: ApiClient = createApiClient(
+      ApiSchema,
+      "https://example.com",
+      global.fetch
+    )
+
+    try {
```
seanim0920, line 42: Had trouble using a mock response from the fetch, it appears as undefined

api-client/aws-fetch.ts
```diff
@@ -0,0 +1,105 @@
...
+  cognitoUser: CognitoUser | null
+}
+
+// refresh cognito token if expired
+async function refreshToken (callback?: () => void): Promise<void> {
+  try {
+    const storedAuthDetails = await AsyncStorage.getItem("cognitoAuth")
+    if (storedAuthDetails) {
+      globalThis.cognitoAuth = JSON.parse(storedAuthDetails)
+    }
```
mhayes853, line 22: Things like auth tokens are probably not best in async storage since they are private credentials, but rather the keychain (or whatever android calls it I'm not sure) since it's encrypted.

[This](https://github.com/oblador/react-native-keychain) looks a like a pretty decent library to facilitate that.

api-client/APIClient.test.ts
mhayes853: You could use something like [msw](https://github.com/mswjs/msw) which is a pretty convenient way to mock network requests without having to mock fetch.

api-client/APIClient.test.ts
```diff
@@ -0,0 +1,81 @@
+import { z } from "zod"
+import { createApiClient } from "./api-client"
+import { GenericApiClient } from "./api-endpoint-types"
+
+const ApiSchema = {
+  "/users/{userId}": {
```
mhayes853, line 6: What if for the schema, we structured it like this:

```ts
{
    users: {
        path: "/users/{userId}",
        get: { /* Zod Schemas */ },
        patch: { /* Zod Schemas */ },
        // ...
    } 
}
```

We could then call into it like:

```ts
const resp = await apiClient.users.get({ /* Stuff */ })
```

I think this reads a lot better and would probably give us better lsp support (and avoids linter warnings).

api-client/api-endpoint-types.ts
```diff
@@ -0,0 +1,44 @@
...
+            (S[K][M] extends { queryParamsType: ZodType<infer Q> }
+              ? { queryParams: Q }
+              : {}) &
+            (S[K][M] extends { pathParamsType: ZodType<infer R> }
+              ? { pathParams: R }
+              : {})
+        ) => Promise<O>
+      }
+    : never
+}
```
mhayes853, line 42: Typescript is truly a work of art 😆

api-client/APIClient.test.ts
seanim0920: Sure, up to @tristan8chu and @nairoldi if they want to give unique names to the endpoints

api-client/index.ts
```diff
@@ -0,0 +1,31 @@
...
+ *   - `pathParams`: An object containing path parameters, if applicable
+ *
+ * Example usage:
+ *   const response = await apiClient["/user"]["GET"]({ queryParams: { id: 1 } });
+ *
+ * @type {ApiClient}
+ */
+export const apiClient: ApiClient = createApiClient(
+  ApiSchema,
+  API_URL ?? "",
```
mhayes853, line 29: Minor nit that could be helpful to others in the future:

Maybe if `API_URL` is undefined we could just throw an error that is more descriptive rather than just throwing in an empty string (an error is fine here because it is strictly a programmer error if it is not present in the env). This could prevent hour long debugging sessions only to find out that you forgot the base url in the env.

mhayes853 (approved):
This is good enough to start for sure. Though I think we should consider the suggestion I had about structuring the schema once we get a chance to play around with this more.

suryatiru (approved):

PR #113, opened 2023-05-04
Event Details Rework

nicoletteL:
This PR mostly updates how the event details component is written by separating segments of code into their own components. This includes a component for adding a friend and one for a profile image and name (and profile image by itself).

I moved the friend state into using a jotai atom as the state needs to be used across components (so far the Toast that displays when adding a friend).

The rest is styling, which includes coloring and sizing updates to the navigation tab bar and the recenter location button so they more closely match the figma designs.


ZoroarkZark (dismissed):

components/bottomTabComponents/AttendanceButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  style?: StyleProp<ViewStyle>
+}
+
+const AttendanceButton = ({ attendeeStatus, style }: LeaveJoinButtonProps) => {
+  return (
+    <View style={[style, styles.bottomTab]}>
+      {isAttendingEvent(attendeeStatus)
+        ? (
+          <OutlinedButton
+            title={attendeeStatus == "hosting" ? "Delete Event" : "Leave Event"}
```
mhayes853, line 18:
```suggestion
            title={attendeeStatus === "hosting" ? "Delete Event" : "Leave Event"}
```

components/profileImageComponents/ProfileImageAndName.tsx
```diff
@@ -0,0 +1,136 @@
...
+                    style={{ marginHorizontal: 8 }}
+                    color={AppStyles.colorOpacity35}
+                    size={6}
+                    name="ellipse"
+                  />
+                  <Headline
+                    style={{ opacity: 0.35 }}
+                    onPress={() => setRequestSent(true)}
+                  >
+                  Friend
```
mhayes853, line 116:
```suggestion
                  Friends
```
The plural sounds more appealing when viewing the UI

lib/NativeMap.ts
```diff
@@ -26,20 +25,26 @@ const mapOptions = (details: EventMapDetails) => {
...
       ? userLocation.location.coordinates
       : undefined
 
   if (!userCoordinates) {
-    await Location.requestForegroundPermissionsAsync()
+    await requestLocationPermissions()
+    userCoordinates =
+      userLocation && userLocation.status === "success"
+        ? userLocation.location.coordinates
+        : undefined
```
mhayes853, line 47: I don't think we need this check. It should be possible to just open maps with destination coordinates and have it figure out the rest.

components/profileImageComponents/ProfileImageAndName.tsx
```diff
@@ -0,0 +1,136 @@
...
+interface ImageAndNameProps {
+  displayFriend?: boolean
+  displayHandle?: boolean
+  style?: StyleProp<ViewStyle>
+  imageStyle?: StyleProp<ImageStyle>
+  username: string
+  userHandle: string
+  eventColor: string
+  toastOffset?: number
+}
```
mhayes853, line 25: Make sure that we can take in the friend status as a type so that this becomes reusable on the attendees list screen. That type could look something like this:

```ts
// Don't know if we need blocked but we definitely need the first 3.
type UserFriendStatus = "not-friends" | "friend-request-pending" | "friends" | "blocked"
```

Also it looks like we're starting to get quite a few flag arguments here, maybe we could have multiple components of the different variants?

components/profileImageComponents/ProfileImageAndName.tsx
nicoletteL: Yeah, you're right.

ZoroarkZark (dismissed):

nicoletteL:
I couldn't figure out how to separate the commits out of the iOS pull request so I just merged the changes into this one.

ZoroarkZark (dismissed):

ZoroarkZark (approved):

mhayes853 (approved):

PR #119, opened 2023-05-21
Fixed package versions for expo 48

seanim0920:


mhayes853:
I still get an issue with android, but it can be fixed by installing `expo-standard-web-crypto` and putting the following in lib/uuid.ts

```ts
import * as _uuid from "uuid"
import { polyfillWebCrypto } from "expo-standard-web-crypto"

polyfillWebCrypto()

/**
 * Generates a v4 UUID string.
 */
export const uuid = () => _uuid.v4()
```

ZoroarkZark (approved):

mhayes853 (approved):

PR #123, opened 2023-05-31
Edit Profile Screen UI

nicoletteL:
This PR contains the UI shown in the designs for the edit profile screen:

- Clickable profile image so integrating image selection should go smoothly
- Working text input that keeps/updates state so backend integration should be smooth (also includes error state on handle)
- Save changes button that disables itself when no changes are made (also when the user enters the same information)
- Close button that warns user when edits have been made but not saved
- Navigation from tab bar -> profile screen -> edit profile screen as well as settings screen
- Navigation headers with functionality
- Feedback from profile screen PR


ZoroarkZark (dismissed):

lib/users/User.ts
```diff
@@ -18,3 +21,39 @@ const IS_VALID_USER_HANDLE_REGEX = /^@[A-Za-z0-9_]{1,15}$/
...
+
+export type UserFriendStatus =
+  | "not-friends"
+  | "friend-request-pending"
+  | "friends"
+  | "blocked"
+
+export type User = {
+  id: string
+  username: string
```
mhayes853, line 33:
```suggestion
  name: string
```

components/common/ConfirmationDialogue.tsx
```diff
@@ -22,7 +22,7 @@ const ConfirmationDialogue = ({ style }: Props) => {
 
   const showActionSheet = () => {
     // To show the Bottom ActionSheet
-    actionSheet.current.show()
+    actionSheet.current!.show()
```
mhayes853, line 25:
```suggestion
    actionSheet.current?.show()
```
Force unwraps cause all sorts of weirdness. (Sorta equivalent to a segfault in C, but JS just gives you undefined instead)

screens/ProfileScreen/Navigation/EditProfileScreenNavWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Returns the current values from a parent `EditProfile` component.
+ */
+export const useEditProfileContext = () => {
+  const context = useContext(EditProfileContext)
+  return { ...context!}
+}
+
+const EditProfileContext =
+  createContext<EditProfileContextValues | undefined>(undefined)
```
mhayes853, line 70: Jotai should obsolete this context

screens/ProfileScreen/Navigation/ProfileScreenNavWrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useEffect (() => {
+    navigation.setOptions({
+      headerTitle: "",
+      headerLeft: () => user.userStatus !== "current-user" && <ChevronBackButton />,
+      headerRight: () => renderHeaderRight(navigation)
+    })
+  }, [navigation])
+
+  return <ProfileScreen user={user}/>
+}
```
mhayes853, line 45: Similar to the edit profile screen navigation, I would also make the same changes here. Though you don't need jotai since there's no shared state.

screens/ProfileScreen/Navigation/ProfileScreensNavigation.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { StackNavigatorType } from "@components/Navigation";
+import { SettingsScreen } from "@screens/SettingsScreen/SettingsScreen";
+import { ProfileScreenNavWrapper } from "./ProfileScreenNavWrapper";
+import { EditProfileScreenNavWrapper } from "./EditProfileScreenNavWrapper";
+
+
+export type ProfileScreensParamsList = {
+  "Profile Screen": ProfileScreenProps
+  "Edit Profile Screen": ProfileScreenProps
+  "Settings Screen": undefined
```
mhayes853, line 12: Stick with camel case names. We've already done it with the reporting screens, and it makes this type look more like a standard typescript type.

screens/ProfileScreen/Navigation/ProfileScreensNavigation.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { ProfileScreenProps } from "../ProfileScreen"
+import { StackNavigatorType } from "@components/Navigation";
+import { SettingsScreen } from "@screens/SettingsScreen/SettingsScreen";
+import { ProfileScreenNavWrapper } from "./ProfileScreenNavWrapper";
+import { EditProfileScreenNavWrapper } from "./EditProfileScreenNavWrapper";
+
+
+export type ProfileScreensParamsList = {
+  "Profile Screen": ProfileScreenProps
+  "Edit Profile Screen": ProfileScreenProps
```
mhayes853, line 11: Also, reminder that we only want an id for whoever's profile it is as the route params, so we can get deep linking. For the case of the current user's profile tab, we can just have a special route that leads to that screen which doesn't need an id prop since we know it's the current user.

stacks/ActivitiesStack.tsx
```diff
@@ -162,9 +176,17 @@ export function TabNavigation () {
...
+        options={({route}) => ({
+            tabBarVisible: getTabBarVisibility(route)
+          })
+        }
+      >
+        {() => 
+          <ProfileStack.Navigator>
+            {profileScreens}
+          </ProfileStack.Navigator>
+        }
```
mhayes853, line 188: Can't we just have a component which holds onto a profile stack navigator instead of this?

ZoroarkZark (approved):

tristan8chu (approved):

screens/ProfileScreen/Navigation/ProfileScreensNavigation.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { AppStyles } from "@lib/AppColorStyle"
+import { EditProfileDismissButton } from "./EditProfileDismissButton"
+import EditProfileView from "../EditProfileScreen/EditProfileView"
+import { EventMocks } from "@lib/events"
+import { useHydrateAtoms } from "jotai/utils"
+import { View } from "react-native"
+import { TouchableIonicon } from "@components/common/Icons"
+
+export type ProfileScreensParamsList = {
+  ProfileScreen: ProfileScreenViewProps
```
seanim0920, line 21: Note: Add linter rule to prevent functions and objects from being in nav props

ZoroarkZark (approved):

nairoldi (approved):

PR #135, opened 2023-07-25
Slim App component

mhayes853:
To begin the mass refactor, the `App` component has been heavily slimmed. It has been dropped from over 300 lines to ~50. Of course, once more app launch logic is required that number will once again rise, but hopefully it does so in a clean and testable manner. However, the biggest benefit of this PR at this time is that there are no longer a sea of warning and error logs when the app launches, so it should make debugging a bit easier.

I also extracted out the `MenuProvider` into a separate component file and renamed `AppQueryClientProvider` to `TiFQueryClientProvider` (I just think that the TiF prefix is more relevant than the App prefix, but tbh this is really a nitty change).

The `App` component itself has also been renamed to `AppView` so that `LiveApp` could just be renamed to `App`. `AppView` is the top most component in the app and as such should follow our headless model. I did omit the `style` prop since there isn't any point in having it for the root of the app, but at the very least it should be consistent in following the separation of logic and view pattern.


App.tsx
```diff
@@ -1,405 +1,49 @@
...
-import { SetDependencyValue } from "./lib/dependencies"
-import { userIdDependencyKey } from "./lib/MiscDependencyKeys"
+import { TabNavigation } from "@stacks/ActivitiesStack"
 import { SafeAreaProvider } from "react-native-safe-area-context"
-
-if (
-  Platform.OS === "android" &&
-  UIManager.setLayoutAnimationEnabledExperimental
-) {
-  UIManager.setLayoutAnimationEnabledExperimental(true)
```
seanim0920, line 79: Need this to allow for layout animations on android https://reactnative.dev/docs/layoutanimation

ZoroarkZark (approved):

PR #138, opened 2023-07-29
Simplify user location logic

mhayes853:
Basically, this gets rid of the unnecessary abstraction around getting the user's location via expo. I also created a `UserLocationFunctions` context such that these functions can be stubbed out in non-production environments, but the main thing is that expo types are now used instead of our own for this logic, as before we just had unnecessary code in performing these weird conversions.

I also renamed `Location` to `TiFLocation` to distinct it from expo's `LocationObject`.


mhayes853:
Seems like the open in maps functionality broke without me catching it earlier... Well I fixed it and event simplified the open in maps code.

ZoroarkZark (approved):

PR #140, opened 2023-08-01
Refactor location search UI

mhayes853:
This PR mainly removed the usage of the dependencies system in the location search feature by making `LocationSearchPicker` a headless component. 

However, it also added a very important rich domain type for aiding in making the actual query which is called `LocationsSearchQuery`. This domain type represents user entered text, and it determines whether or not to load from the user's recent locations, or AWS. It also makes the code a bit more cohesive in the UI for showing elements related to the recent locations, as we don't just check a random empty string in the UI anymore, rather that happens in the domain type.


mhayes853:
Since this is the "UI refactor", I also decided to add a little icon next to each search result to make the screen much less devoid of life. It also makes distinguishing recent location search results much easier:
IMG_2962

screens/LocationSearch/Icon.tsx
```diff
@@ -0,0 +1,50 @@
...
+export type LocationSearchIconProps = {
+  backgroundColor: string
+  name: IoniconName
+  style?: StyleProp<ViewStyle>
+} & AccessibilityProps
+
+/**
+ * The icon next to each search result in the location search feature.
+ */
+export const LocationSearchIconView = ({
```
seanim0920, line 21: Should we use a common icon component

seanim0920 (approved):

PR #160, opened 2023-09-16
Update change password

mhayes853:
This PR updates the change password to match the rest of the auth screens. Additionally, there were some bugs in the change password that needed to be fixed which required that we also need to display multiple error messages to the user.

Furthermore, the `isFocused` field was causing some weird issues on iOS when trying to do the "focus the next field when the user taps next on their keyboard" so we're back to using refs. This of course involved wrapping every single text field component with `forwardRef`...

Additionally, the text field error now animates in when it appears, which makes layout animations possible within these forms (as seen in the new change password screen).

I also added a storybook story for the screen and removed the password re-enter since it was an extra step that isn't really needed.

### DIFF
--- a/.storybook/.ondevice/Storybook.tsx
+++ b/.storybook/.ondevice/Storybook.tsx
@@ -34,6 +34,9 @@ import SignUpMeta, {
 import VerifcationCodeMeta, {
   Basic as VerifcationCodeBasic
 } from "../components/VerificationCode/VerifyCode.stories"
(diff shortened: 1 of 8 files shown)


ZoroarkZark (commented):
Forward ref stuff seems somewhat dangerous, might need second opinion on whether or not it looks safe for use

ZoroarkZark (approved):

PR #164, opened 2023-09-21
Secure Auth Storage

ZoroarkZark:
For this pull request, I have created a storage system that uses ExpoSecureStore to properly store information instead of AsyncStorage. 

I have tested this so far, and from the tests, it works as intended.  Please let me know if there is anything that might strengthen these cases, or could help my methods.


lib/SecureStorage.ts
```diff
@@ -0,0 +1,72 @@
+import * as ExpoSecureStore from "expo-secure-store"
+
+export type SecureStorage = typeof ExpoSecureStore
```
mhayes853, line 3:
```suggestion
export interface SecureStore {
  getItemAsync (key: string): Promise<string>
  deleteItemAsync (key: string): Promise<void>
  setItemAsync (key: string): Promise<void>
}
```

This is so that you don't need to mock all the irrelevant methods (like `isAvailiableAsync`) for testing.

lib/SecureStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.store = store
+    this.keyList = []
+  }
+
+  setItem (key: string, value: string) {
+    if (!this.keyList.find((k) => k === key)) {
+      this.keyList.push(key)
+    }
+    this.cache.set(SECURE_STORAGE_KEY_PREFIX + key, value)
+    this.store.setItemAsync("KeyList", JSON.stringify(this.keyList))
```
mhayes853, line 23: Nit: I would extract a constant for `"Keylist"` since you use it in different places.

lib/SecureStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+          this.cache.set(SECURE_STORAGE_KEY_PREFIX + k, storageValue)
+        }
+      })
+    )
+  }
+}
+
+// Amplify.configure({
+//   storage: AmplifySecureStorage
+// })
```
mhayes853, line 72:
```suggestion
```

tests/helpers/SecureStorage.ts
```diff
@@ -0,0 +1,84 @@
...
+  ) {
+    this.TestAmplifyStore.set(key, value)
+  }
+
+  async getItemAsync (key: string) {
+    let result = this.TestAmplifyStore.get(key) || null
+    if (!result) {
+      result = null
+    }
+    return result
```
mhayes853, line 74:
```suggestion
    return this.TestAmplifyStore.get(key) ?? null
```

mhayes853 (commented):
Pretty good for the most part, just some minor adjustments and you'll be fine.

mhayes853 (approved):

PR #183, opened 2023-10-31
Search Client Module Logic

ZoroarkZark:
* This pull request is the logic behind a search client for locations; Geo does not currently work for the module, although mocks that simulates the function that uses it do exist.

* For now, tests have been provided to make sure the logic works; the ones using the function that uses Geo have been commented out for now.


location-search/SearchClient.test.ts
```diff
@@ -0,0 +1,157 @@
...
+          isRecentLocation: true
+        },
+        {
+          location: location1,
+          annotation: "attended-event",
+          isRecentLocation: true
+        }
+      ])
+    })
+    it("gives an empty array when called with faulty or nonexistent information", async () => {
```
ZoroarkZark, line 96: returns, not gives

location-search/SearchClient.test.ts
```diff
@@ -0,0 +1,157 @@
...
+import { asyncStorageSaveRecentLocation } from "./RecentsStorage"
+import {
+  awsPlaceToTifLocation,
+  searchRecentLocations,
+  searchWithRecentAnnotations
+} from "./SearchClient"
+import { mockLocationSearchFunction } from "./mocks"
+import { LocationsSearchQuery } from "./models"
+
+Geo.configure(awsconfig)
```
seanim0920, line 14: We can remove this part since you mention it’s causing issues. I’ll test it myself at a later date

location-search/SearchClient.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        coordinates: mockLocation3,
+        placemark: { name: "New York" }
+      }
+
+      const testQuery = new LocationsSearchQuery("New York")
+
+      await asyncStorageSaveRecentLocation(mockTiFLocation1, "attended-event")
+      await asyncStorageSaveRecentLocation(mockTiFLocation2, "hosted-event")
+      await asyncStorageSaveRecentLocation(mockTiFLocation3, "hosted-event")
+
```
seanim0920, line 125: Is the test updated to combine the data from aws + asyncstorage

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * Uses either {@link searchRecentLocations}, if querying for the user's current recent locations OR
+ * uses {@link searchWithRecentAnnotations}, along with the given query and center, if querying for
+ * locations in a different location.
+ *
+ * @param query A {@link LocationsSearchQuery} given for the search function to use.
+ * @param center An optional {@link LocationCoordinate2D} that designates where the location search is occurring from.
+ * @returns A Promise of an array of {@link LocationSearchResult}s, or usable data given from the search client.
+ */
+export const useLocationSearch = async (
```
mhayes853, line 25: This isn't a hook, so don't use the `use` moniker.

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+  query: LocationsSearchQuery,
+  center?: LocationCoordinate2D
+): Promise<LocationSearchResult[]> => {
+  if (query.sourceType === "user-recents") return await searchRecentLocations()
+  else {
+    return await searchWithRecentAnnotations(
+      query,
+      mockLocationSearchFunction,
```
mhayes853, line 33:
```suggestion
      awsGeoSearchLocations,
```

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+ * @param center An optional {@link LocationCoordinate2D} given for the search to center in where it should be searching.
+ * @returns A Promise of an array of {@link TiFLocation}s.
+ */
+export const awsGeoSearchLocations = async (
+  query: LocationsSearchQuery,
+  center?: LocationCoordinate2D
+) => {
+  const options = {
+    maxResults: 10,
+    biasPosition: center ? [center?.longitude, center?.latitude] : undefined
```
mhayes853, line 79: I don't think you need to `?` on center since ts should infer that it's not undefined.

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+  const asyncRecentSearchResults =
+    await asyncStorageLoadSpecificRecentLocations(searchCoordinates)
+  const filteredResults = ArrayUtils.compactMap(
+    convertedRemoteSearchResults,
+    (remoteSearchPoint) => {
+      if (
+        !asyncRecentSearchResults.find(
+          (asyncPoint) =>
+            JSON.stringify(asyncPoint.location) ===
+            JSON.stringify(remoteSearchPoint.location)
```
mhayes853, line 135: I would extract a `areCoordinatesEqual` function in the same file where the coordinates type comes from. This is just an unnecessary JSON conversion.

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+ * @returns A Promise of an array of {@link LocationSearchResult}s.
+ */
+
+export const searchWithRecentAnnotations = async (
+  query: LocationsSearchQuery,
+  searchFunction: (
+    query: LocationsSearchQuery,
+    center?: LocationCoordinate2D
+  ) => Promise<TiFLocation[]>,
+  center?: LocationCoordinate2D
```
mhayes853, line 114: Nit: would move this above the `searchFunction` parameter.

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+): Promise<LocationSearchResult[]> => {
+  const remoteSearchResults = await searchFunction(query, center)
+  const convertedRemoteSearchResults = remoteSearchResults.map((point) => ({
+    location: point
+  }))
```
mhayes853, line 119: What is the point of this map?

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+  console.log(
+    "Converted Remote Search: " + JSON.stringify(convertedRemoteSearchResults)
+  )
```
mhayes853, line 122: Remove this

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+          (asyncPoint) =>
+            JSON.stringify(asyncPoint.location) ===
+            JSON.stringify(remoteSearchPoint.location)
+        )
+      ) {
+        return remoteSearchPoint
+      }
+    }
+  )
+  const mergedResults = asyncRecentSearchResults.concat(filteredResults)
```
mhayes853, line 142: This changes the ordering of the results, which we should avoid and leave to `searchFunction`. It's that function's job to to ensure the correct ordering which may order based on relevancy for instance. Doing this destroys that ordering and possibly provides a worse UX.

location-search/SearchClient.ts
```diff
@@ -0,0 +1,148 @@
...
+  return mergedResults.map((result) => ({
+    location: result.location,
+    annotation: result.annotation,
+    isRecentLocation: !!result.annotation
+  }))
```
mhayes853, line 147: I think what I would do here is map on the `remoteSearchResults`, then do the `.find` stuff you did earlier. This would preserve the ordering whilst still merging the data properly.

location-search/SearchClient.test.ts
```diff
@@ -0,0 +1,141 @@
...
+          annotation: "hosted-event",
+          isRecentLocation: true
+        },
+        {
+          location: mockTiFLocation1,
+          annotation: "attended-event",
+          isRecentLocation: true
+        }
+      ])
+    })
```
mhayes853, line 139: I think you can add 2 more results to this assertion:
1. A result without an annotation but is a recent location.
2. A result that isn't a recent location and thus has no annotation.

location-search/SearchClient.ts
ZoroarkZark: It's meant to convert the information from the remoteSearchResults (which give TiFLocations) to objects that contain that information labeled as "location", so that it would be easier to concatenate the two arrays

location-search/SearchClient.ts
ZoroarkZark: Can't, center is an optional parameter

lib/location/Location.ts
```diff
@@ -24,6 +24,19 @@ export type LocationCoordinate2D = Readonly<
   z.infer<typeof LocationCoordinates2DSchema>
 >
 
+/**
+ * Checks for equality on two sets of coordinates.
+ */
+export const coordinateMatch = (
```
ZoroarkZark, line 30: coordinateMatch can be taken multiple meanings, name should specify that it is checking if coords are equal

location-search/SearchClient.ts
ZoroarkZark: Instead, could make center 
```center: LocationCoordinate2D | undefined```

mhayes853 (approved):

PR #190, opened 2023-11-18
Upgrade AWS Amplify auth

mhayes853:
Upgrades AWS amplify auth to v6, which is a huge major version upgrade that replaces pretty much every function. This PR breaks a decent chunk of our existing auth code, so it will remain in draft until all is well and good.


auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
-  setItem (key: string, value: string) {
-    const chunks = stringToChunks(value)
-    this.keyChunkMappings.set(key, chunks.length)
-    this.saveKeyChunkMappings()
-    this.cache.set(key, value)
-    for (let i = 0; i < chunks.length; i++) {
-      this.store.setItemAsync(secureStoreKey(key, i), chunks[i])
+  async removeItem (key: string) {
+    const keyChunkMappings = await this.loadKeyChunkMappings()
+    if (keyChunkMappings) {
```
mhayes853, line 52: I think `loadKeyChunkMappings` should be written such that it never returns undefined so that you don't have to do this dance in each one of the get/set/remove methods.

auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
+    const keyChunkMappings = await this.loadKeyChunkMappings()
+    if (keyChunkMappings.size > 0) {
+      const amountOfChunks = keyChunkMappings.get(key) ?? 0
+      const chunks = [] as string[]
+      for (let i = 0; i < amountOfChunks; i++) {
+        const chunk = await this.store.getItemAsync(this.secureStoreKey(key, i))
+        if (chunk) {
+          chunks.push(chunk)
+        }
+      }
```
mhayes853, line 89: This can also be done with `Promise.allSettled` to avoid doing the async work sequentially.

auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
-  getItem (key: string) {
-    return this.cache.get(key)
+  async setItem (key: string, value: string) {
+    const keyChunkMappings = await this.loadKeyChunkMappings()
+    if (keyChunkMappings) {
+      const chunks = this.stringToChunks(value)
+      for (let i = 0; i < chunks.length; i++) {
+        const chunkKey = this.secureStoreKey(key, i)
+        await this.store.setItemAsync(chunkKey, chunks[i])
+      }
```
mhayes853, line 72: Use `Promise.allSettled` to avoid doing this work sequentially.

auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
     await this.saveKeyChunkMappings()
-    this.cache = new Map()
   }
 
-  async sync () {
-    if (this.syncPromise) {
-      return await this.syncPromise
+  private async loadKeyChunkMappings () {
+    if (this.keyChunkMappingsPromise) {
+      return await this.keyChunkMappingsPromise
```
mhayes853, line 118: I'm pretty sure that this would load outdated chunks and thus cause the get/set/remove methods to have incorrect behavior. There should be another check to see if we have `this.keyChunkMappings` as not undefined and just return that in that case. The reason I don't think you've done this however is because there's no test cases that expose this flaw, which would be doable by first doing some operations in `testStorage1` and then doing some operations (and not just calling `getItem`) in `testStorage2`.

auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
-      KEY_CHUNK_MAPPINGS_KEY,
-      JSON.stringify(Array.from(this.keyChunkMappings))
-    )
+  stringToChunks = (data: string) => {
+    const numChunks = Math.ceil(data.length / STORE_CHUNK_SIZE)
+    const chunks = [] as string[]
+    for (let i = 0; i < numChunks; i++) {
+      chunks.push(data.slice(i * STORE_CHUNK_SIZE, (i + 1) * STORE_CHUNK_SIZE))
+    }
+    return chunks
```
mhayes853, line 150: This should be private.

auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
-}
 
-const stringToChunks = (data: string) => {
-  const numChunks = Math.ceil(data.length / STORE_CHUNK_SIZE)
-  const chunks = [] as string[]
-  for (let i = 0; i < numChunks; i++) {
-    chunks.push(data.slice(i * STORE_CHUNK_SIZE, (i + 1) * STORE_CHUNK_SIZE))
+  secureStoreKey = (name: string, chunkIndex: number) => {
+    const key = SECURE_STORAGE_KEY_PREFIX + name + `___chunk${chunkIndex}`
+    return key
```
mhayes853, line 155: This should be private.

auth/CognitoSecureStorage.ts
```diff
@@ -36,109 +37,121 @@ const STORE_CHUNK_SIZE = 2048
...
+    const keyChunkMappings = await this.loadKeyChunkMappings()
     await Promise.allSettled(
-      Array.from(this.keyChunkMappings).map(async ([key, amountOfChunks]) => {
+      Array.from(keyChunkMappings).map(async ([key, amountOfChunks]) => {
         await Promise.allSettled(
           ArrayUtils.repeatElements(amountOfChunks, async (chunkIndex) => {
-            await this.store.deleteItemAsync(secureStoreKey(key, chunkIndex))
+            await this.store.deleteItemAsync(
+              this.secureStoreKey(key, chunkIndex)
+            )
```
mhayes853, line 107: I think you may just be able to call `removeItem` here.

mhayes853 (commented):
I think there need to be more tests specifically for `removeItem` and around doing operations in 2 store instances (which is supposed to simulate the user launching the app in which a new instance of the class would be created).

seanim0920:
[Auth: Upgrade Cognito*](https://trello.com/c/g33qRwDw/333-auth-upgrade-cognito)

PR #193, opened 2023-11-24
Arrivals: EventArrivalsTracker

mhayes853:
This PR begins the series of PRs for the event arrival tracking functionality. The first in this series being of course laying out the core of the feature which is embedded in the `EventArrivalsTracker` class.

This class effectively ensures that all arrivals that are being tracked are also being geofenced, and it will serve as the core class that components in future PRs will interact with. It depends on 2 things, an instance of `PendingEventArrivalNotifications`, which has an implementation class using Async Storage, and a function to effectively start geocoding all the coordinates of the upcoming event arrivals.

This PR for now doesn't contain anything on the refresh policy or any actual geofencing atm.


mhayes853:
Since the server is now gonna handle the scheduling of the notifications as discussed, I changed many of the names to more concise versions to reflect that. I also removed most of the properties on `EventArrival` since most of them existed for the sake of notification scheduling.

seanim0920 (approved):

seanim0920 (approved):

suryatiru (approved):

PR #197, opened 2023-11-25
AuthNavigation stack + folder restructure

ZoroarkZark:
* Adjusting navigation structure (removing ActivitiesStack is one of the first things done), as well as adding an AuthNavigationStack
* At the moment, Auth is plugged into the navigation. Not sure if any other implementation will be used once the next update rolls around.


root-feature/Navigation/AuthNavigation.tsx
```diff
@@ -0,0 +1,75 @@
...
+  SignUpParamsList &
+  ForgotPasswordParamsList
+
+const Stack = createStackNavigator<AuthNavigationStackParamList>()
+
+const tiFAPI = new TiFAPI(
+  createAWSTiFAPIFetch(
+    new URL("https://623qsegfb9.execute-api.us-west-2.amazonaws.com/prod/")
+  )
+)
```
mhayes853, line 34: We could add a static property for this so one could just use `TiFAPI.production` or something like that. Also, I think there's an environment variable for the production URL called `API_URL` which you can access via:

```ts
import { API_URL } from "env"
```

root-feature/Navigation/TabNavigation.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const tabHiddenRoutes = [
+    "EditProfileScreen",
+    "SettingsScreen",
+    "ChangePasswordScreen"
+  ]
+  if (tabHiddenRoutes.includes(routeName)) {
+    return false
+  }
+  return true
+}
```
mhayes853, line 20: We don't need this, we'll put the screens that don't use the tab bar in the root stack navigator.

components/bottomTabComponents/BottomNavTabBar.tsx
mhayes853, line 1: Move this component into the bottom tab navigation file in the root feature.

root-feature/AppView.tsx
```diff
@@ -0,0 +1,27 @@
...
+  isFontsLoaded: boolean
+}
+
+export const AppView = ({ isFontsLoaded }: AppProps) => {
+  if (!isFontsLoaded) return null // TODO: - Splash Screen?
+  return (
+    <NavigationContainer>
+      <Stack.Navigator>
+        <Stack.Screen
+          name="Activities Screen"
```
mhayes853, line 18:
```suggestion
          name="tabs"
```

mhayes853 (approved):

PR #202, opened 2023-11-30
Reorganize some modules and the Lib directory

mhayes853:
This is a pretty hefty PR that incorporates many of the folder structure changes we've talked about recently. In no particular order (excluding the lib folder changes):

- Adds a new `date-helpers` module
- Adds a new `location` module (which now houses `location-search`)
- Adds a new `test-helpers` module (finally)
- Adds a new `content-reporting` module (for reporting events/users)
- Removes the `hooks` folder

Now for the lib directory. We'll define a feature/module in the lib folder as a feature/module that is small enough to fit in 1 file. A good example of this are both analytics and haptics (the non-native code part), while both are significant features, they also don't have enough area to constitute a separate module like `date-time`. Another good example is for containing configuration/wrapper code for external dependencies.

Additionally, there is now a `utils` directory within `lib` which contains extensions to the standard library (eg. `StringUtils`, `ArrayUtils`, and now `MathUtils`), and or bits of functionality extending the capabilities of 3rd party libraries (eg. `ZodUtils`, `AsyncStorageUtils`). 

You may notice that some  "library files" (files in `lib` that are named after a library like `ReactQuery`) are in the root of `lib` whilst others are in `lib/utils`. The reason for this is because the "library files" in the utils folder contain generic algorithms that are built on top of the existing library, and are not particular to our app in anyway. In theory, these files can be used completely independently in another project. This contrasts the root of `lib` which contains files that _are_ geared towards our app and our architecture, and that should be noticeable that some of these files may contain type/variable/function/etc. names that have "TiF" in them.

I decided to still leave `lib/users` and `lib/events` in for now, as I have no idea what to do with them in the forseable future, though I'm sure we'll find a resolution for when the time comes.

The root `tests` folder is also starting to get quite small. I haven't yet modularized the explore and the edit-events features yet, especially since edit-events is due for a redesign, but once we get back to them we will modularize them.

I think overall, we're doing a decent job with vertically slicing. I definitely don't want the app to end up with >30 root level modules since it can get "confusing", though for me I actually prefer this 🥹.


mhayes853:
Ok, to summarize the madness from earlier, it seems that CI had issues resolving the Models file in the arrivals feature, so I renamed to something completely random which fixed it. Now, apparently renaming it back to Models again somehow fixed the issue 😆.

ZoroarkZark (approved):

PR #210, opened 2023-12-19
Event Arrivals: API Integration

mhayes853:
Integrates the existing arrivals code with the backend. I will keep this in draft as the endpoints are still in progress on the backend side of things.

There were also 2 significant "refactorings" needed to conform to new behavior that we discussed, and to prevent circular imports:
1. In `EventArrivalsOpertationResult`, `"non-upcoming"` has now been renamed to the more general `"remove-from-tracking"` as there are other reasons to remove the event from tracking other that its time being shifted back such as the event being deleted.
2. I made a new `@shared-models` module, which currently houses some of the location types and `EventArrivalsOperationResult`, as the `@api-client` module needs them and doing it this way avoids circular import errors.


mhayes853:
We decided on using a new common currency type that groups event ids by region rather than managing a flat array of `EventArrival` instances due to it being easier to geofence efficiently. However, this also required some pretty substantial changes to the `EventArrivalsTracker` due to the new behavior this and the updated responses from the endpoints entailed.

seanim0920 (approved):

PR #216, opened 2024-01-01
Event Arrival Banner

mhayes853:
The banner UI for when the user arrives at an event. Overall it's a pretty simple piece of UI, but the exact banner messages did take some calculus to concoct (so I naturally wrote a good amount of tests for this). I did so by simply calculating each sentence independently, and then combining the together at the end. I think this approach should be fine when we localize since each distinct sentence could be represented by a localized string key, and ordering of sentences doesn't change in other languages AFAIK.


mhayes853:
I've reverted the hook commit for now. I think we'll leave that functionality up to the full details implementation next sprint so I know better what to do.

seanim0920 (approved):

PR #230, opened 2024-01-13
UseLoadEventDetails

mhayes853:
Implements a hook to load the event for the event details screen. Since the details screen needs to support multiple cases, I there's naturally cases for when the user is blocked by the host, when the event has been deleted, or when it's not found. We do show some info to the user when the event is blocked which can be found in the `BlockedEvent` type.


seanim0920 (approved):

event-details/EventDetailsLoading.tsx
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * A result from loading a single event for the details screen.
+ */
+export type EventDetailsLoadingResult =
+  | { status: "not-found" | "deleted" }
+  | { status: "blocked"; event: BlockedEvent }
+  | { status: "success"; event: CurrentUserEvent }
+
+export type UseLoadEventDetailsResult =
+  | { status: "loading" | "not-found" | "deleted" }
```
seanim0920, line 17: May want to use this type for other queries, but we can wait until more integration

mhayes853: For sure

PR #273, opened 2024-03-05
Event Details: Misc/Menu Actions

mhayes853:
Dropdown/Breadcrumb Menu UI for the event details. It currently supports 6 distinct actions, each being available for users with different event roles (Hosting, Attending, Not-Participating).

- Reporting the event (Attendee, Non-Participant).
- Copying the event (All roles).
- Contacting host (Attendee, Non-Participant).
- Editing the event (Host).
- Sharing the event (All roles).
- Inviting friends (Host, Attendee).
- Assigning a new host (Host).

I used `react-native-menu` for the code itself to display the native dropdown menu on each platform. Currently, this library only supports using SF Symbols on iOS, so we'll have to do something when we need to design our custom icons at some point.

For the event `MenuAction`s from `react-native-menu`, I put them in a giant `EVENT_MENU_ACTIONS` config object using the `satisfies` keyword. `satisfies` allows us to get autocomplete on the properties of `MenuAction` from `react-native-menu` on the `EVENT_MENU_ACTIONS` object when we need to add a new/or update existing actions. I then wrote a function called `formatEventMenuActions` which fills in the `MenuAction`s with data from the event. Additionally, on iOS the menu will display the actions in the reverse order that you pass them into `MenuVIew`, but not on Android. Therefore, I added a call to `Array.reverse` on Android in `formatEventMenuActions`, and listed the actions in reverse by default for each property in `ATTENDEE_STATUS_ACTIONS`.

For the share sheet, I added a prop that asynchronously loads the `ShareContent` since we haven't yet decided on what that content will exactly be.


android/app/src/main/java/com/tifapp/FitnessApp/MainApplication.kt
```diff
@@ -45,9 +46,9 @@ class MainApplication : Application(), ReactApplication {
   override fun onCreate() {
     super.onCreate()
     SoLoader.init(this, false)
-    if (!BuildConfig.REACT_NATIVE_UNSTABLE_USE_RUNTIME_SCHEDULER_ALWAYS) {
```
seanim0920, line 48: expo prebuild failure

event-details/Menu.tsx
```diff
@@ -0,0 +1,184 @@
...
+export type EventMenuAction =
+  (typeof EVENT_MENU_ACTION)[keyof typeof EVENT_MENU_ACTION]
+
+export type EventMenuActionID = EventMenuAction["id"]
+
+/**
+ * Given a list of {@link EventMenuAction}, formats appropriate content from
+ * `event` in place of `"%@"` for any of the given menu actions.
+ */
+export const formatEventMenuActions = (
```
seanim0920, line 89: function name is somewhat vague

seanim0920 (dismissed):

seanim0920 (approved):

seanim0920 (dismissed):

seanim0920 (approved):

PR #276, opened 2024-03-11
Event Details: Replace Edit Menu Action with Block Toggling

mhayes853:
Replaces the "Edit Event" menu action with a new action to block/unblock the host since we display the edit event action next to the host row on the details screen.

I made some changes to the `EVENT_MENU_ACTIONS` config object which allow the title and image to be computed from a function so that creating dynamic content was easier.

### DIFF
--- a/.storybook/components/EventDetails/EventDetails.stories.tsx
+++ b/.storybook/components/EventDetails/EventDetails.stories.tsx
@@ -17,7 +17,6 @@ import {
 import { mockPlacemark } from "@location/MockData"
 import { EventCountdownView, eventCountdown } from "@event-details/Countdown"
 import { dateRange, dayjs, now } from "@date-time"
-import { View } from "react-native"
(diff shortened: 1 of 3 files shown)


.storybook/components/EventDetails/EventDetails.stories.tsx
```diff
@@ -30,6 +29,8 @@ import { createTestQueryClient } from "@test-helpers/ReactQuery"
 import { QueryClientProvider } from "@tanstack/react-query"
 import { EventDetailsMenuView } from "@event-details/Menu"
 import { View } from "react-native"
+import { NonCompliantChecksCount } from "aws-sdk/clients/iot"
```
mhayes853, line 32:
```suggestion
import { NonCompliantChecksCount } from "aws-sdk/clients/iot"
```

seanim0920 (approved):

PR #287, opened 2024-03-25
Explore Events: Integration

mhayes853:
Integrates the events-by-region endpoint with the explore events screen. For now, I've left out the part where the event arrivals tracker needs to be updated, and we can do that in a follow up.


api-client/client.ts
```diff
@@ -181,7 +181,7 @@ const parseResponseBody = async (
 ) => {
   try {
     return await responseSchema.parseAsync(json)
-  } catch {
+  } catch (e) {
```
seanim0920, line 184: Do you want to print this error?

seanim0920 (approved):

api-client/client.ts
mhayes853: I can make a PR in the shared repo for this.

suryatiru (approved):

PR #289, opened 2024-04-03
Update Module Structure

mhayes853:
Our module structure is way too adhoc, and this sometimes allows modules to be importing other modules that they shouldn't be importing (eg. `event-details` importing `location-search`). That being said, there seems to be a general pattern with how we're importing things that works, so this PR just codifies that and fixes some of the violations of the code.

Essentially there are now 3 kinds of modules:
- `core-root`
  - Contains the entry point view of the application, and is allowed to import anything it needs since it wires the whole app together.
  - ***Nothing*** is allowed to import `core-root` other than `App.tsx` and `AppRegistery.js`, or any storybook story.
- `core` modules
  - Non-generic modules for each end-to-end feature of the application.
  - Have their names prefixed with `core-`.
  - Can import any module other than `core` modules.
    - This prevents the event details from depending on the user profile and vice versa, causing an import cycle.
- Everything else
  - `lib` or primarily shared modules that multiple core modules may need.
  - Can import _any_ module **except for `core-root`**.

In other words, the plan is to separate the non-generic domain-specific-core parts of the app from the generic parts of the app. With this separation, there would be no need for situations where location search depends on explore events, since the code in explore events is geared specifically towards explore events. If location search and explore events need to share something, then either a generic module can be created with the shared code, or the shared code can be put inside `lib` if it's small enough.

There currently 5 `core` modules with more coming soon:
- `core-event-details`
- `core-auth`
- `core-explore-events`
- `core-location-search`
- `core-content-reporting`

I also broke `arrival-tracking` into its own module, since it was pretty distinct from the event details. I didn't make it a core module since the interfaces it provides can easily be consumed by multiple core modules if necessary.


mhayes853:
Changed the naming from `core` to `boundary` based on our discussions. However, I'm still keeping the root module as `core-root` to differentiate it from the other modules.

seanim0920 (approved):

suryatiru (approved):

PR #290, opened 2024-04-03
Sqlite: Fallback to in-memory database when opening fails

mhayes853:
Opening a SQLite database can fail for a variety of reasons. Some include:
- Permissions issues
- Lack of disk space
- Path not found
- etc.

In our case, we use SQLite mainly for storing client-side data. Since that data is not necessarily integral to the app it wouldn't make sense to alert the user about any of the above opening failure cases. Therefore, the best approach is to fail silently. For this, I've made it so that if the call to `openDatabaseAsync` fails for any reason, I fallback to calling `openDatabaseAsync` with an _in memory_ path. This way, sqlite operations can proceed somewhat normally without being a deterrence to the user experience.

`TiFSQLite` now accepts a new argument in its constructor which performs the opening logic to which the constructor calls with the provided `path` argument. If opening the `path` argument fails, it recalls the new open argument with `SQLITE_IN_MEMORY_PATH` which should succeed nearly always.

Alongside this, I also made some minor refactorings to the internals of the `TiFSQLite` class.

### DIFF
--- a/lib/SQLite.ts
+++ b/lib/SQLite.ts
@@ -73,8 +73,7 @@ export class TiFSQLite {
 
   private async runQueueUntilEmpty() {
     while (!this.isQueueEmpty) {
-      const transaction = this.queuedTransactions[0]
-      await transaction()
+      await this.queuedTransactions[0]()
       this.queuedTransactions.shift()
     }
   }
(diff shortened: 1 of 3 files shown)


seanim0920 (approved):
Needs to-do comment to add trace logs when falling back to in-memory store

PR #293, opened 2024-04-16
Implement EventAttendeesList class

ZoroarkZark:
For this PR, I wanted to implement the `EventAttendeesList` class in order to not only make the process of making lists of attendees easier in the code, but also implement properties that cement a list as having a host, total attendee count, and an array of attendees.

An issue I was running into for the original implementation of the `AttendeesList` was that there was not a set standard for what the host should be in a list of attendees, so it was implemented in-line from the return value. However, if we were to change this host implementation in the future, it would be more trouble to change that one aspect of the code, or it might be forgotten after other modules get developed. Therefore,  the solution was to create a class of a list of attendees. This would allow for a standard to be developed for the host, and for the total attendee count, or the attendees of the list, to be easily accessed, and for operations directly onto the list to be performed.

To start, I had created an `EventAttendeesList` class, and had given it the `host` and `totalAttendeeCount` properties. There is also an included `attendees` function, in order to obtain a flat list of attendees from the current pages, if a function needs to use a simple array of attendees.

After this, I remembered that the `hostPicker` in another branch would require the use of a local update to the attendees list, when an attendee is removed or swapped. Therefore, I had begun to implement `swapHost`, which would swap the current host with the attendee specified, which is required for changing who the host is. In addition, I added `removeAttendee`, for locally removing an attendee from the attendee list.

I had created some tests for not only the new `EventAttendeesList` class that was made, but also updated the ones for the original implementation for the attendees list, as instead of returning properties, the hook `useAttendeesList` would be returning an instance of `EventAttendeesList` directly. This definitely needed updating, due to the new class.


event-details-boundary/attendees-list/AttendeesList.tsx
```diff
@@ -124,36 +117,7 @@ export type AttendeesListViewProps = {
...
-          </View>
-        </View>
-      }
-      data={attendees}
-      onRefresh={refresh}
-      keyExtractor={(item) => `attendee-${item.id}`}
-      renderItem={({ item }) => renderAttendee(item)}
-      onEndReached={fetchNextGroup}
-    />
-  )
```
mhayes853, line 158: Was there any reason for removing this?

event-details-boundary/attendees-list/EventAttendeesList.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        EventAttendeeMocks.BlobJr.id
+      )
+      expect(attendeesList.host).toEqual(EventAttendeeMocks.BlobJr)
+      expect(attendeesList.attendees()).toEqual([
+        EventAttendeeMocks.AnnaAttendee,
+        EventAttendeeMocks.Alivs,
+        EventAttendeeMocks.BlobSr
+      ])
+    })
+    it("switches host with attendee right after original host in list", () => {
```
mhayes853, line 45: Shouldn't these test names mention that page of the attendee we're swapping the host with? 1 test seems like it was meant for swapping the host with an attendee on the first page, and the other the second page. "list" and "original list" was a bit confusing, so I think "first page" and "second page" would work better.

mhayes853 (commented):
Base this off of `shared-integration`, some imports still need to be cleaned up (this should also fix the CI error).

event-details-boundary/attendees-list/AttendeesList.tsx
ZoroarkZark: No; was removed when unused, I believe

event-details-boundary/attendees-list/AttendeesList.tsx
```diff
@@ -148,11 +142,11 @@ export const AttendeesListView = ({
           {ListHeaderComponent}
           <View>
             <Headline> Attendees </Headline>
-            <Headline> ({totalAttendeeCount})</Headline>
+            <Headline> ({attendeesList.totalAttendeeCount})</Headline>
           </View>
         </View>
       }
-      data={attendees}
+      data={attendeesList.attendees()}
```
mhayes853, line 149:
```suggestion
      data={useMemo(() => attendeesList.attendees(), [attendeesList])}
```

mhayes853 (approved):

PR #298, opened 2024-04-24
Settings: Genericize Settings Store

mhayes853:
We have multiple kinds of settings that are stored differently in the app (Local and API Settings), but the code to observe them in the UI is heavily shared. So this PR introduces a new `PersistentSettingsStore` class, `SettingsStorage` interface, and `SettingsStore` interface, and updates the local settings to use this new system.

The `SettingsStore` interface abstracts over the act of observing the settings. I made it an interface since alongside `PersistentSettingsStore` I'm also working on another implementation of `SettingsStore` which will also sync settings to the API. `SettingsStore` has 3 requirements, which include the standard subscribe, update, and get current value functionality needed to use this with `useSyncExternalStore`.

The `SettingsStorage` interface abstracts over the persistence of settings. Mainly, we can have 2 implementations, 1 for local settings, and 1 for User settings which both use different tables in SQLite. It has both loading and saving requirements, and an additional `tag` requirement that is used for error logging. The tag must uniquely identify the instance.

The `PersistentSettingsStore` class implements the `SettingsStore` interface, and adds persistence to the settings whenever the values change. To do this, one must pass an instance of `SettingsStorage` and a value containing all the default settings into the constructor. The class also swallows persistence errors thrown by `SettingsStorage` and logs them at a `warn` level.

Lastly, I updated the local settings to use this new system by writing a `SQLiteLocalSettingsStorage` class which conforms to `SettingsStorage`. Additionally, I'm currently working on implementing the remote user settings using this new system.


seanim0920 (approved):

seanim0920:
[Settings Store - Front-End](https://trello.com/c/COVMVPrL/629-settings-store-front-end)

PR #299, opened 2024-04-25
User Settings Sync

mhayes853:
Adds user settings synchronization with the API by building on top of the classes and interfaces introduced in #298. For this PR description, when I refer to local settings I do not mean the type `LocalSettings`, but rather the local persistence of the `UserSettings` type.

A few things to note before getting into the sync process:
1. Exponential Backoff/Retry behavior is handled particularly by `tanstack/query-core`. In other words instead of writing a hook and using `useQuery` and `useMutation`, I used `QueryClient` and `MutationObserver` directly (see `addAPIUserSettingsExponentialBackoff`).
2. I decided to make a new implementation of `SettingsStore` called `UserSettingsSynchronizingStore`, as the approach I took looked quite weird when trying to put all the sync logic in a hook. This class uses `PersistentSettingsStore` under the hood for the subscribing and local persistence logic.
3. I made another conformance to `SettingsStorage` which again uses SQLite and stores the settings inside a single-row table called "UserSettings".

## Sync Process

For now, I think a simple `version` number check is fine, and would work better than a last-write wins date scheme since device clocks can get out of sync easily. Essentially, if the local settings version is the same or greater than the API settings version, then we push the local settings to the API and store the API settings locally otherwise. I've also looked briefly at [Vector Clocks](https://en.wikipedia.org/wiki/Vector_clock), though I don't think this is necessary atm.

This check is performed in `userSettingsRefreshAction` which will return `"push-local"` if the local settings need to be pushed. While `"push-local"` is technically the only return status that is directly used by `userSettingsRefreshAction`, I still decided to return specific status strings ("pull-remote"`, etc.) for all the other cases to serve as code documentation for what each scenario in the function represents.

## `UserSettingsSynchronizingStore`

This class implements `SettingsStore` and adds the sync behavior on top of a `PersistentSettingsStore` instance.

When `update` is called, I made sure to debounce saving the settings to the API since the user may change many settings in quick succession. The `flushPendingChanges` method can be used to manually push the debounce changes to the API at any time, such as when the app is backgrounded.

`refresh` should be called when the application foregrounds, or when the user switches logs in/out of their account. This method will fetch the settings from the API, and run `userSettingsRefreshAction` to decide whether or not to push the local settings or save the API settings. The method will also invalidate any pending debounce update.

## Retry Behavior

Saving changes to the API comes with exponential backoff, and automatic retry behavior when the network comes back online. I was able the leverage `MutationObserver` from react-query to achieve this since it has all of this behavior built-in. `MutationObserver` is the class that `useMutation` uses under the hood, but since we're not in a hook the class needs to be used directly. The class itself takes the same parameters that `useMutation` needs with the addition of having to pass a required `QueryClient` since there's no react context to get the current `QueryClient`. Like `useMutation`, the class has a `mutate` method which performs the mutation.


seanim0920:
[github-trello integration](https://trello.com/c/y6zY4kQW/545-github-trello-integration)

suryatiru (dismissed):

tifapp (approved):

seanim0920 (approved):

seanim0920:
[Settings Sync - Front-End interface](https://trello.com/c/WR2VGO1r/630-settings-sync-front-end-interface)

PR #315, opened 2024-05-08
Settings: Privacy Settings + Initial Settings Components

mhayes853:
1. Creates a new `settings-boundary` module.
2. Creates an initial screen for configuring privacy settings.
3. Adds a few reusable components for the upcoming settings screens in the `components` sub-module of `settings-boundary`

This PR will remain in draft until I hear back on the exact language that should be used for the descriptions on the screen.

I also re-ran `pod install` on the iOS side.

The first section of toggles relates to the privacy settings we've discussed, and the second section lists all the permissions that the app needs and is using. 

When the user presses a toggle on the permissions section, 1 of 2 things can happen:
1. If the permissions is not granted and we can ask for it, then the app will ask for the permission explicitly.
2. Open system settings otherwise.

Here's what it looks like atm (the system permissions prompt seems to be omitted from screen recording on iOS, but rest-assured it did appear).
[vid](https://imgur.com/a/0UpR4FS)

TASK_5vj63JXY


tifapp:
Linked Ticket: https://trello.com/c/5vj63JXY

settings-boundary/PrivacySettings.tsx
```diff
@@ -0,0 +1,198 @@
...
+    <ScrollView style={style} contentContainerStyle={styles.contentContainer}>
+      <PreabmleSectionView />
+      <LearnMoreSectionView onPrivacyPolicyTapped={onPrivacyPolicyTapped} />
+      <ShareSectionView />
+      <PermissionsSectionView permissions={permissions} />
+    </ScrollView>
+  )
+}
+
+const PreabmleSectionView = () => (
```
ZoroarkZark, line 83: *Preamble

suryatiru (approved):

PR #318, opened 2024-05-09
Settings: Support Simple Arrays

mhayes853:
Some settings values like event duration presets and notification triggers are best represented by arrays of primitive JS values. So this adds another union case to the `SettingValue` type to include support for a `SettingValue[]`. I also updated `areSettingsEqual` to ensure that arrays are compared for equality by value and not by reference.

TASK_UVOItWdN


tifapp:
Linked Ticket: https://trello.com/c/UVOItWdN

mhayes853:
Had to remove the deep array value equality check, it was causing some re-render issues when working on the other settings screens.

seanim0920 (approved):

PR #319, opened 2024-05-11
Settings: Notification Settings

mhayes853:
Implements the Notification Settings screen alongside https://github.com/tifapp/TiFShared/pull/31. This will remain in draft until the wording on the screen has been finalized.

I had to update the user settings database table to support the notification triggers, an array of strings. Since there isn't any relational need for the triggers, I decided to store them in a TEXT column by serializing them CSV style.

The screen itself has 2 forms. Firstly, if the user hasn't enabled notification permissions, a section at the top will alert them of this and provide a button to enable notification permissions. This button behaves exactly like one of the permissions toggles in #315. Additionally, the rest of the screen is greyed out and unresponsive until notification permissions are enabled.
IMG_0071\

The second form is the screen but not greyed out. There are currently 2 sections, an events notification section, and profile notification section. Of course, we will have to add more to this soon, but this is fine for now.
IMG_0072

On the technical side, I added a couple more settings helper views and adjusted existing ones to support section-level disabling. I also wrote a helper function to make writing selectors easier called `settingsSelector`. `settingsSelctor` has a variadic parameter where you pass keys to a `Settings` object, and it returns a function that selects the settings values from a `Settings` object instance using those keys.

https://trello.com/c/kJRwtAGS


mhayes853:
UI Updates based on feedback.
IMG_0078
IMG_0079
IMG_0081 2

suryatiru (approved):

PR #327, opened 2024-05-24
Logs + Main SQLite DB separation

ZoroarkZark:
## Problem/Purpose

The purpose of this PR is to separate the logs from the main tables of the SQLite database, where settings, location placemarks, etc. are stored. This is done in order to easily create a copy of them for importing purposes, or to access the current ones in a more isolated database. 

## Implementation Steps

1. In order to adjust the implementation of the database, we needed to adjust the TiFSQLite class.
2. Going there, we remove its `migrateV1` function in favour of a more generic `migrate` parameter, so that we can specify what gets run for the database.
3. We add a namespace with two functions: `main`, which makes most tables, and `logs`, which makes only the logs table. 
4. We adjust any implementations of `TiFSQLite` or `testSQLite` accordingly.

### DIFF
--- a/lib/SQLite.ts
+++ b/lib/SQLite.ts
@@ -290,7 +290,7 @@ export namespace Migrations {
       id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
       label TEXT NOT NULL,
       message TEXT NOT NULL,
-      level TEXT NOT NULL CHECK(level IN ('debug', 'info', 'warn', 'error')),
+      level TEXT NOT NULL CHECK(level IN ('debug', 'info', 'warn', 'error', 'trace')),
       stringifiedMetadata TEXT,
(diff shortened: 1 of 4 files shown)


lib/SQLite.ts
```diff
@@ -287,3 +208,92 @@ export class ExpoSQLExecutable implements SQLExecutable {
...
+      `
+    ])
+  }
+  export const logs = async (db: SQLExecutable) => {
+    await db.run`
+    CREATE TABLE IF NOT EXISTS Logs (
+      id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+      label TEXT NOT NULL,
+      message TEXT NOT NULL,
+      level TEXT NOT NULL CHECK(level IN ('debug', 'info', 'warn', 'error')),
```
mhayes853, line 293: Make sure to add `trace` to this.

mhayes853 (approved):

PR #329, opened 2024-05-26
Settings: Fix iOS Unblock Success Banner Bug

mhayes853:
`react-native-root-toast` will sometimes still be displayed even if the `visible` prop is false. Since the toast state is controlled by an optional `BlockListUser`, if you tapped unblock on another user at the right time, this will happen:

IMG_0097

AFAIK, this only happened on iOS, but the fix was to use the `useLastDefinedValue` utility hook that we have. This way, even if `mostRecentlyUnblockedUser` is undefined, it will ensure that a valid name is displayed on the banner.

https://trello.com/c/bHPmsdXH


tifapp:
Linked Ticket: https://trello.com/c/bHPmsdXH

seanim0920 (approved):

seanim0920:
It would be good to add a comment to that line with the explanation from the PR.

PR #330, opened 2024-05-28
Settings: Calendar Settings + Update Components

mhayes853:
Implements the calendar settings screen, which is by far the simplest settings screen. Due to our discussions, there is only 1 setting on this screen for now, the day of the week to start the calendar. It has quite a simple look too:
Simulator Screenshot - iPhone X - 2024-05-28 at 02 31 37

For the menu, I made a `SettingsMenuPickerView` to display the menu and the selected option. The selected option has a checkmark in the menu and has its title displayed to the right of "Start Week On". There is also an interesting-looking icon next to the title, this is to mimic the look of iOS's native dropdown picker menu. For comparison:

IMG_0100

Unfortunately, `react-native-menu` does not provide any functionality to have the checkmark on the left-hand side like in native Swift code, so I settled with no icon and a checkmark on the right side to compensate.

## Why not put this in general settings with only 1 setting?

2 Reasons.
1. The calendar will be a highly visible part of the app, so it should have an easily accessible screen to customize the appearance in settings.
2. It's likely we'll add more settings to the calendar once we actually build it.

https://trello.com/c/0pcdwFrp


settings-boundary/CalendarSettings.tsx
```diff
@@ -0,0 +1,50 @@
...
+export type CalendarSettingsProps = {
+  style?: StyleProp<ViewStyle>
+}
+
+export const CalendarSettingsView = ({ style }: CalendarSettingsProps) => (
+  <SettingsScrollView style={style}>
+    <WeekdayPickerSection />
+  </SettingsScrollView>
+)
+
```
mhayes853, line 19:
```suggestion
export type CalendarSettingsProps = {
  style?: StyleProp<ViewStyle>
}

export const CalendarSettingsView = ({ style }: CalendarSettingsProps) => (
  <SettingsScrollView style={style}>
    <StartWeekOnSectionView />
  </SettingsScrollView>
)

```

seanim0920 (dismissed):

seanim0920 (approved):

PR #332, opened 2024-05-28
Settings: Root

mhayes853:
Implements the main settings screen (minus navigation). Overall, the screen is straightforward.
IMG_0101

For the account info link, I made sure that it doesn't appear when the user is not signed in. To do this, I updated the hooks inside `UserSession.tsx` and used the `IfAuthenticated` component to decide whether or not to render the link. The `UserSessionProvider` now expects a function to return the current `UserSession` rather than a boolean that says whether or not the user is signed in.

We'll probably add a lot more to the screen in the near future, and we could possibly shrink the amount of padding between sections (this would go for all settings screens).

https://trello.com/c/oyp2MSvn


tifapp:
Linked Ticket: https://trello.com/c/oyp2MSvn

seanim0920:
Looks clean and inviting. Nice design.

settings-boundary/RootSettings.tsx
```diff
@@ -0,0 +1,179 @@
...
+  areNotificationsEnabled
+}: MainScreenLinksSectionProps) => (
+  <SettingsCardSectionView>
+    <SettingsNavigationLinkView
+      iconName="settings"
+      iconBackgroundColor={AppStyles.black}
+      title="General"
+      onTapped={() => onSettingsScreenLinkTapped("generalSettings")}
+    />
+    <IfAuthenticated
```
seanim0920, line 54: Nice that we can easily toggle visibility of the screens like this

settings-boundary/RootSettings.tsx
```diff
@@ -0,0 +1,179 @@
...
+      title="Events"
+      onTapped={() => onSettingsScreenLinkTapped("eventSettings")}
+    />
+    <SettingsNavigationLinkView
+      iconName="notifications"
+      iconBackgroundColor={AppStyles.yellow}
+      title="Notifications"
+      rightAccessory={
+        <BodyText style={styles.notificationsText}>
+          {areNotificationsEnabled ? "On" : "Off"}
```
seanim0920, line 82: Can we call it "areSystemNotificationsEnabled", I was unclear at first which notification this was referring to

event-details-boundary/Menu.test.tsx
```diff
@@ -188,6 +197,10 @@ describe("EventDetailsMenu tests", () => {
       })
     })
 
+    const ensureUserIsSignedIn = () => {
```
seanim0920, line 200: nit - can we call it something like "useSignedInUser", "ensure" sounds like we're checking a condition

seanim0920 (approved):

PR #334, opened 2024-05-29
Organize Native Code

mhayes853:
Soon, some more native code will need to be written, this PR renames the `TiFTravelEstimates` swift package into `TiFNative` so that it can hold *any* native code we want. Furthermore, I've moved all the travel estimates code into a subfolder within that package.

https://trello.com/c/DqNjdYvm


tifapp:
Linked Ticket: https://trello.com/c/DqNjdYvm

PR #336, opened 2024-06-07
Roswaal: Test Progress Monitoring

mhayes853:
See https://github.com/tifapp/AcceptanceTestBot/pull/3

Implements the client-side javascript needed to interact with the automated acceptance test tool. 2 main classes are used by the generated code: `RoswaalClient` and `RoswaalTestCase`. You can find more specifics in their doc comments. Generally, these classes shouldn't be used directly, and should only be used by the generated code.

## `RoswaalTestCase`

This class represents a single test case, and it has 2 methods for adding actions and running the test. The latter will run each appended action sequentially until all are finished, or until one fails, and return a progress report. If an action fails, then all subsequent actions are considered failures.

## `RoswaalClient`

This class is responsible for holding onto the progress of all test cases, and has a method to upload the progress on all test cases to the tool. This `upload` method is called in a global jest teardown hook.

A `RoswaalClient` can be constructed with or without an `uploadURL`. The latter case is useful for running the tests locally when no upload is needed. When calling `upload`, a warning message will be logged instead of an upload occurring.

To support uploads in a CI environment, you'll set the `ROSWAAL_UPLOAD_RESULTS_URL` used as the `uploadURL` of the singleton `roswaalClient`. This URL doesn't exist yet, but it will contain a security token as a query parameter to ensure that only trusted clients can upload progress results.

## New NPM Commands

I have modified `package.json` to support running the acceptance tests, roswaal infrastructure tests, and the main frontend tests in isolation. For an overview:
- `npm run test` runs the frontend test suite.
- `npm run test-roswaal` runs the tests for the roswaal infrastructure (just `RoswaalClient` and `RoswaalTestCase` for now).
- `npm run test-acceptance` runs the acceptance tests with detox. You'll need to pass a `--configuration` flag to this command indicating which device you want to test on.

## Generated Test Example

Copied from https://github.com/tifapp/AcceptanceTestBot/pull/3

```ts
// Generated by Roswaal, do not touch.
import * as TestActions from "./TestActions"
import { launchApp } from "../Launch"
import { RoswaalTestCase } from "../TestCase"
import { roswaalClient } from "../Client"

test("A", async () => {
  const testCase = new RoswaalTestCase("A", TestActions.beforeLaunch)
  // Johnny is signed in
  testCase.appendAction(TestActions.ensureJohnnyIsSignedIntoHisAccount)
  // Set Location to Oakland
  testCase.appendAction(TestActions.setLocationToOakland)
  await roswaalClient.run(testCase)
})
```

## Tickets

https://trello.com/c/UH6YeaPS


tifapp:
Linked Ticket: https://trello.com/c/UH6YeaPS

seanim0920 (approved):

PR #338, opened 2024-06-11
Fix EAS Build: Move TiFShared import to root + point TiFShared version to main branch

seanim0920:
The linter moved the TiFShared import to the middle of the file which caused build issues. I'll move the TiFShared declaration to the root so we prevent that from accidentally happening.

https://trello.com/c/jgUBLFjj/674-android-eas-build-fix-keeps-crashing


tifapp:
Linked Ticket: https://trello.com/c/jgUBLFjj

mhayes853 (approved):

PR #339, opened 2024-06-11
Add watchmanconfig to try and improve expo startup time

seanim0920:
Ignoring "node_modules" should reduce the amount of files being tracked when expo starts up
https://facebook.github.io/watchman/docs/config

TASK_UNTRACKED


mhayes853 (approved):
Works for me, but does this resolve the waiting issue that you and @ZoroarkZark were having?

suryatiru (approved):

PR #342, opened 2024-06-19
Mail Composer Wrapper

mhayes853:
The help and support screen (and possibly future screens) present a success alert when the user sends an email successfully through the mail composer. Unfortunately, Android does not give access to this information, but expo still decides it's a good idea to return `MailComposerStatus.SENT` even if the user did not send an email. Due to this, it's possible for a "We received your feedback!" alert to be presented even when no email was sent on Android.

There are 2 possible fixes for this:
1. Check that both the status is `MailComposerStatus.SENT` and that `Platform.OS === "android"` every time we need to evaluate the result of a mail composition session.
2. Wrap `expo-mail-composer`'s `composeAsync` function and return a custom `EmailCompositionResult` union type based on the check from fix 1.

I went with fix 2.

## Tickets

https://trello.com/c/IGbGWAuV

### DIFF
--- a/lib/EmailComposition.ts
+++ b/lib/EmailComposition.ts
@@ -1,0 +1,34 @@
+import {
+  MailComposerOptions,
+  MailComposerStatus,
+  composeAsync
+} from "expo-mail-composer"
+import { Platform } from "react-native"
+
+export type EmailCompositionResult =
+  | "success"
+  | "cancelled"
+  | "non-determinable"
+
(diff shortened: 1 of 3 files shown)


tifapp:
Linked Ticket: https://trello.com/c/IGbGWAuV

tifapp:
Linked Ticket: https://trello.com/c/IGbGWAuV

ZoroarkZark (approved):

PR #351, opened 2024-08-03
More Flexible Circular Ionicon Component

ZoroarkZark:
## Problem/Purpose

This pull request aims to allow the circular ionicon component, used in many screens, including SignUp, EventSettingsScreen, and others that have an ionicon with a circular background.

## Proposed Solution

In order to perform this, there has been a size prop included with the circular ionicon that the component uses to determine its size, rather than the original default size (24) that it used. For this to work, the original formulas were altered, and additional CSS properties (justify-content and align-items) were added in order to keep the ionicon looking centered and flexible.

## Results

Ideally, this should allow any future screen that uses a circular ionicon to set the size to whatever is needed, rather than sticking to a default size that might not fit a particular screen.

## Tickets

https://trello.com/c/dwgQ8U4V


components/common/Icons.tsx
```diff
@@ -144,6 +145,7 @@ export const CircularIonicon = ({
   backgroundColor,
   name,
   style,
+  size,
```
mhayes853, line 148: Default this to `DEFAULT_ICON_SIZE`

components/common/Icons.tsx
```diff
@@ -153,41 +155,38 @@ export const CircularIonicon = ({
           circularStyles.iconBackground,
           {
             backgroundColor,
-            width: DEFAULT_ICON_SIZE * 1.5,
-            height: DEFAULT_ICON_SIZE * 1.5
+            width: size * 1.5,
+            height: size * 1.5
```
mhayes853, line 159:
```suggestion
            width: size * useFontScale({ maximumFontSizeMultiplier: maximumFontScaleFactor }) * 1.5,
            height: size * useFontScale({ maximumFontSizeMultiplier: maximumFontScaleFactor }) *  1.5
```
Without this, you get this when the user uses an accessibility font size.

IMG_0151

mhayes853 (commented):
Sorry for the later review, you just need to ensure that it works with larger/accessibility font sizes. I've added the code to make this work in suggestions.

mhayes853 (commented):
Sorry for the later review, you just need to ensure that it works with larger/accessibility font sizes. I've added the code to make this work in suggestions.

mhayes853 (commented):
Sorry for the later review, you just need to ensure that it works with larger/accessibility font sizes. I've added the code to make this work in suggestions.

mhayes853 (commented):
This is a comment to make github behave.

mhayes853 (approved):

PR #377, opened 2024-08-20
Compile Logs

mhayes853:
Implements log compilation for the help and support settings screen. This now means the user can email us their locally stored logs when submitting a bug report.

This PR introduces a native code module called `tif-fs` for native filesystem operations.

## Difficulty with `expo-sqlite` File Paths

Sqlite is just a file, so it should just be easy enough to zip up right? Normally this is the case, but unfortunately, I had to resort to hacks to get the URL to the logs sqlite database in the first place. This is because `expo-sqlite` does not have an API to return the path to an open database. Instead, you create a database with a name, and `expo-sqlite` manages a `SQLite` directory under the hood. For instance:
```ts
const db = await SQLite.openDatabaseAsync("foo.db")
```
We don't specify where to store `foo.db`, expo stores it in a special folder, and never exposes an API to retrieve the path.

The solution required looking at the source code of `expo-sqlite`, and writing a native code to retrieve the path to the logs database using the same methods as `expo-sqlite`.

## Zipping

There exists a library called `react-native-zip-archive`, but it would require us to substantially increase our iOS deployment target. Therefore, I wrote the zip function in native code using native Swift and Kotlin libraries.

## Tickets

https://trello.com/c/2yeKCEfS


tifapp:
Linked Ticket: https://trello.com/c/2yeKCEfS

seanim0920 (approved):

PR #379, opened 2024-08-24
Region Monitoring Harness: Improvements

mhayes853:
Adds background support to the region monitoring harness. This will be in draft as we debug the accuracy of the tracking.

## Tickets

https://trello.com/c/28WJIizX

PR #381, opened 2024-08-25
Region Monitoring Demo Build - Added minimap + language changes for testers

seanim0920:
To get the ios version to build in expo, I updated the packages from expo doctor's recommendations, deleted the ios folder and then injected the TiFNative dependency into the podfile. Potentially doing a prebuild with the updated packages could fix the expo issue.

image


mhayes853:
#379

PR #382, opened 2024-08-26
Event Arrivals: Deduplicate Arrival Operations

mhayes853:
There was a complex infinite loop that occurred with the `EventArrivalsTracker` which would cause the background task to run repeatedly, and spam requests to the backend. For brevity, here is the description of the loop from slack.

Screenshot 2024-08-25 at 7 24 55 PM

The fix is to deduplicate arrival operations when the updated `hasArrived` matches the stored `hasArrived`. I added a new method requirement to the `EventArrivalsStorage` to return the `hasArrived` value for a particular stored region, and then called this inside `handleGeofencingEvent` in `EventArrivalsTracker`.

## Tickets

https://trello.com/c/28WJIizX

### DIFF
--- a/arrival-tracking/Tracker.test.ts
+++ b/arrival-tracking/Tracker.test.ts
@@ -14,8 +14,7 @@ import { repeatElements } from "TiFShared/lib/Array"
 import {
   addTestArrivals,
   expectOrderInsensitiveEventArrivals,
-  regionWithArrivalData,
-  removeTestArrivals
+  regionWithArrivalData
 } from "./TestHelpers"
 import { EventArrivals } from "./Arrivals"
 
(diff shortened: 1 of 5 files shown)


tifapp:
Linked Ticket: https://trello.com/c/28WJIizX

arrival-tracking/Tracker.test.ts
```diff
@@ -186,6 +175,15 @@ describe("EventArrivalsTracker tests", () => {
     expect(callback).toHaveBeenCalledTimes(1)
   })
 
+  it("should only perform the arrivals operatio when arrival status changes for region", async () => {
```
seanim0920, line 178: typo

seanim0920 (approved):

PR #384, opened 2024-09-02
Edit Event: Durations Slider

mhayes853:
Implements a custom slider for the picking the duration of an event based on how the Structured app does it. 

https://github.com/user-attachments/assets/b505d982-bf0b-4f50-a3a8-53abedc9c1db

Under the hood, I got to learn about utilizing React Native Gesture Handler and Reanimated to the fullest which was quite fun.

One weird design decision is that the picker takes a `durationAtom` prop instead of a pair of `duration` and `onDurationChanged` props. This is because the former prop offers better performance (ie. It reduced lag on Android devices) than the traditional pair of props since I can scope rerenders to just the slider knob.

Essentially, the main approach was to let react native and CSS perform the layout, then read that layout information, and then base the current `duration` value off the slider knob's current position. I held the layout information as a state variable of `Map<number, LayoutRectangle | undefined>` where each key is a duration option. The logic for manipulating this map was complex enough to warrant some unit tests IMO.

Lastly, if the user has < 3 presets set, then the slider will add a few default presets such that the slider is not empty.

## Tickets 

https://trello.com/c/KH4nzJzH


tifapp:
Linked Ticket: https://trello.com/c/KH4nzJzH

.storybook/components/EditEvent/DurationPicker.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+            presetOptions={presets}
+          />
+        </Provider>
+        <Button
+          title="Filter"
+          onPress={() => {
+            if (presets !== PRESETS) {
+              setPresets(PRESETS)
+            } else {
+              setPresets((p) => p.filter((i) => i % 2 !== 0))
```
ZoroarkZark, line 40: So this filters out the even indexed presets, I'm assuming? Is this mostly for displaying bigger sizes of the sections?

.storybook/components/EditEvent/DurationPicker.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export default EditEventDurationsMeta
+
+const PRESETS = [
+  300 * 15,
+  601 * 15,
+  900 * 15,
+  1201 * 15,
+  3601 * 15,
+  5400 * 15
```
ZoroarkZark, line 19: Why are these multiplied by 15? If they're presets, wouldn't they be multiplied by a factor of 60 to represent minutes or hour intervals?

edit-event/DurationPicker.tsx
```diff
@@ -1,9 +1,399 @@
...
+        marginTop: withTiFDefaultSpring(!isSliding.value ? INNER_TRACK_GAP : 0),
+        marginLeft: withTiFDefaultSpring(
+          !isSliding.value ? INNER_TRACK_GAP : 0
+        ),
+        borderBottomLeftRadius: withTiFDefaultSpring(
+          isSliding.value ? INNER_TRACK_GAP * 2 : INNER_TRACK_GAP
+        ),
+        borderTopLeftRadius: withTiFDefaultSpring(
+          isSliding.value ? INNER_TRACK_GAP * 2 : INNER_TRACK_GAP
+        )
```
ZoroarkZark, line 109: So with these, since they use `INNER_TRACK_GAP` often, why was that used instead of having the plain results from multiplication?
Would it be for versatility, since `INNER_TRACK_GAP` could be changed at some point?

edit-event/DurationPicker.tsx
```diff
@@ -1,9 +1,399 @@
...
+      )
+      if (nextDuration !== duration) runOnJS(setDuration)(nextDuration)
+    })
+    .onFinalize(() => {
+      const nextPosition =
+        sliderPosition.value < 0
+          ? 0
+          : layoutForDuration(stateEntries, duration)?.x ?? 0
+      sliderPosition.value = withTiFDefaultSpring(nextPosition)
+      runOnJS(setDuration)(durationAtPosition(stateEntries, nextPosition))
```
ZoroarkZark, line 196: How is runOnJS being used here? From what I could tell from research, it calls `setDuration` asynchronously, but I'm not quite sure how this would work.
From what I assume, it's to update the current duration (`duration`) if it doesn't match the `nextDuration` given by the picker itself, on `onUpdate.` For `onFinalize`, I'm not sure.

.storybook/components/EditEvent/DurationPicker.stories.tsx
mhayes853: This was just for testing what it would look like with 3 durations in storybook.

.storybook/components/EditEvent/DurationPicker.stories.tsx
mhayes853: This was another test in storybook, this time to ensure that the text size of the formatted durations fit nicely in the slider.

edit-event/DurationPicker.tsx
mhayes853: Yes

But also because it wasn't obvious _why_ I had to multiply by the padding if I just made it a magic number. If you look at the code for some of the functions that use `EditEventDurationPickerState`, you'll also see use of the content. By naming this constant, it's obvious that the padding refers to the space between the inner track (The inner black background) and the outer track.

edit-event/DurationPicker.tsx
mhayes853: Reanimated doesn't let you call any given function inside the code for the gesture handler since it runs on a separate thread controlled by native code. JavaScript is a single threaded language, so react native runs the JS on a dedicated thread. This creates a problem for multithreaded code like in this case.

You'll notice that some functions have the string `"worklet"` at the start. This string tells reanimated that this function can be called on the native thread, and the reanimated babel plugin will do the work of making that compatible. `setDuration` is a function that is owned by jotai, so we cannot mark it as `"worklet"`. Therefore, we can get around this with `runOnJs`, which will run `setDuration` on react native's JS thread making everyone happy.

ZoroarkZark (approved):

ZoroarkZark (approved):

PR #390, opened 2024-09-20
Event settings native duration picker

seanim0920:
To allow users to set their event durations on android, this PR adds the time picker to the duration screen.

Depends on 
https://github.com/tifapp/TiFShared/pull/56 
https://github.com/tifapp/FitnessProject/pull/392

ios:
image

Different sizes on android
image
image
image

Follow up:
Allow font, text format and highlight style to be configurable from react native

https://trello.com/c/PLcfUIei/728-event-settings-native-duration-picker-kotlin


tifapp:
Linked Ticket: https://trello.com/c/PLcfUIei

android/app/build.gradle
mhayes853, line 1: Don't touch this file.

modules/event-settings-wheel-duration-picker/src/TimePickerModule.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { StyleProp, ViewProps, ViewStyle } from "react-native"
+import { NativeViewGestureHandler } from "react-native-gesture-handler"
+
+type OnDurationChangeEvent = {
+  duration: number
+}
+
+export type TimePickerProps = {
+  initialDuration: number
+  onDurationChange: (event: { nativeEvent: OnDurationChangeEvent }) => void
```
mhayes853, line 12: Do we need to expose the underlying `nativeEvent` here? I think it would be better to wrap this up, and just use the duration value directly as the function parameter.

modules/event-settings-wheel-duration-picker/src/TimePickerModule.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <NativeViewGestureHandler disallowInterruption>
+      <NativeView
+        style={style}
+        onDurationChange={onDurationChange}
+        {...props}
+      />
+    </NativeViewGestureHandler>
+  )
+}
```
mhayes853, line 33: On iOS we don't want to use `requireNativeViewManager` because we'll just use the built-in duration picker component instead.

modules/event-settings-wheel-duration-picker/android/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+
+android {
+  namespace "expo.modules.timepicker"
+  buildFeatures {
+    compose true
+  }
+  // Kotlin compiler version must be tied to Kotlin version
+  // according: https://developer.android.com/jetpack/androidx/releases/compose-kotlin
+  composeOptions {
+    kotlinCompilerExtensionVersion = "1.5.10"
```
mhayes853, line 43:
```suggestion
    kotlinCompilerExtensionVersion = "1.5.14"
```

mhayes853 (commented):
Remove the iOS native code from this, as we'll just use `@react-native-community/datetimepicker` to render the native duration picker.

mhayes853:
See #395

PR #397, opened 2024-10-02
Edit Event: Form

mhayes853:
The form for editing events. This will remain in draft until all the design quirks are worked out.

The general approach to the form uses jotai atoms to ensure that only certain parts of the form are rendered when values change. I think it was easier to do things this way than to use `react-hook-form` like previously as the tools provided by that library are not hard to write from scratch (about ~90 lines of code from pure jotai/zod), and we also get to avoid documentation digging. Therefore, I will remove `react-hook-form` as a dependency in a future PR.

## Tickets

https://trello.com/c/nQs1JJgL
https://trello.com/c/4I9AsAG3


tifapp:
Linked Ticket: https://trello.com/c/4I9AsAG3

ZoroarkZark (approved):

PR #398, opened 2024-10-02
Tifapi integration

seanim0920:
## Problem/Purpose
To make integrations smoother and ensure the backend API types are kept in sync with the frontend, we can add a common API type schema that both projects can refer to, so the frontend always makes valid requests to the API, and the backend always returns valid responses. This PR covers the frontend api integration.

## Implementation Steps
Aside from replacing the TiF API constructor, most of the changes were made to tests, using the new mockTiFServer and mockTiFEndpoint helpers over direct msw instances.

https://trello.com/c/HCMYSMx5/734-tifshared-api-schema-front-end-integration


tifapp:
Linked Ticket: https://trello.com/c/HCMYSMx5

event-details-boundary/Details.test.tsx
```diff
@@ -192,20 +192,14 @@ describe("EventDetailsLoading tests", () => {
   describe("LoadEventDetails tests", () => {
     fakeTimers()
 
-    it("should return canceled when a 204 occurs", async () => {
```
seanim0920, line 195: Currently backend returns event details regardless of when the event ended, so we should revisit this in the future

arrival-tracking/ArrivalsOperation.test.ts
```diff
@@ -23,19 +22,17 @@ describe("ArrivalsOperation tests", () => {
...
-    }) => {
-      expect(await request.json()).toEqual(TEST_REGION)
-      return HttpResponse.json({ trackableRegions: EXPECTED_ARRIVAL_REGIONS })
-    }
-
     test("arrived", async () => {
-      mswServer.use(
-        http.post(TiFAPI.testPath("/event/arrived"), testBodyHandler)
-      )
+      mockTiFServer({
```
seanim0920, line 26: extract to separate function

mhayes853 (approved):

PR #400, opened 2024-10-04
Standardize Alerts

mhayes853:
Whenever we present alerts on a screen, we typically do it through an `XXX_ALERTS` constant object where each value of the object is an object that contains the title, message, and buttons of the alert. This pattern works well for testing since we can keep all the alert contents in the constant object so that we don't have to update the tests when values change.

However, the shape of these alerts objects does not play well with `Alert.alert`, which sometimes caused awkward scenarios where we didn't assert the contents of alerts (because it was too annoying!) and other non-standard alert patterns in the codebase. 

This PR introduces a wrapper `presentAlert` function that presents one of these alert objects. In addition, there are now standard types and jest matchers specific to our pattern of handling alerts, so I expect everything to converge to this pattern from now on.

Here's an example of the new standardized pattern:
```ts
// In SomeFeature.ts

export const SOME_FEATURE_ALERTS = {
  thingHappened: {
    title: "Thing Happened",
    description: "That's too bad!"
  },
  // We make sure that the function is an optional parameter as it makes it so that we don't have to pass a function
  // when trying to use one of the alert matchers in testing.
  confirmThing: (confirm?: () => void) => ({
    title: "Confirm",
    description: "You want to destroy reality?"
    buttons: [{ text: "Yes", onPress: confirm }, { text: "No" }]
  })
} satisfies AlertsObject // This gets us autocomplete for alert properties whilst retaining the constant object

const foo = () => {
  presentAlert(SOME_FEATURE_ALERTS.confirmThing(() => console.log("Thing Confirmed")))
}

// In SomeFeature.test.ts

const { alertPresentationSpy, tapAlertButton } = captureAlerts()

it("should ask for confirmation", async () => {
  foo()
  // The `onPress` function of buttons are omitted when comparing alerts, so we don't need to pass a function to confirmThing.
  expect(alertPresentationSpy).toHaveBeenPresentedWith(SOME_FEATURE_ALERTS.confirmThing())
  await tapAlertButton("Yes") // This will still invoke the action on the actual alert
  // ...
})
```

In addition to `toHaveBeenPresentedWith`, there's also a `toHaveBeenNthPresentedWith` matcher that functions like `toHaveBeenNthCalledWith` but for alerts.

Overall, I decided to keep `alertPresentationSpy` because we can still assert useful things that are typical spy-related things such as how many calls have been presented. However, make sure that you only use these new matchers in conjunction with `alertPresentationSpy`.

## Tickets

https://trello.com/c/lA1qQ4mB


tifapp:
Linked Ticket: https://trello.com/c/lA1qQ4mB

seanim0920 (approved):

PR #401, opened 2024-10-08
Common Bottom Sheet Component

mhayes853:
Everytime we needed to do a bottom sheet, it was really tedious setting up the ref, passing the shared values, etc. Therefore, I decided to make a common component that we can use everywhere. Do note that you will need to wrap the child views of the sheet with `BottomSheetView` or `BottomSheetFlatlist` if you're rendering a list.

The sheet component presentation is state-driven, meaning we do not use a ref to present the sheet. Instead, you use the `isPresented` or the `item` props, the former is a boolean that checks for presentation, and the latter takes an optional value. If the value is not undefined, then the sheet will be presented.

## Tickets

https://trello.com/c/lA1qQ4mB


event-details-boundary/JoinEvent.tsx
```diff
@@ -13,14 +13,13 @@ import {
...
-import React, { useEffect, useRef, useState } from "react"
-import { Alert, StyleProp, StyleSheet, View, ViewStyle } from "react-native"
-import { TouchableIonicon } from "@components/common/Icons"
+import React, { useState } from "react"
+import { StyleProp, StyleSheet, View, ViewStyle } from "react-native"
+import { IoniconCloseButton } from "@components/common/Icons"
 import { BodyText, Title } from "@components/Text"
-import { useSafeAreaInsets } from "react-native-safe-area-context"
-import { useSharedValue } from "react-native-reanimated"
+import { useSafeAreaInsets } from "react-native-safe-area-context""
```
mhayes853, line 20:
```suggestion
import { useSafeAreaInsets } from "react-native-safe-area-context"
```

ZoroarkZark (approved):

PR #405, opened 2024-10-08
Blue Tinted Colors

mhayes853:
Adjusts the primary colors to become more blue tinted. I think this change makes the UI look slightly more fun. Here's a before and after (top before, bottom after).
IMG_0190
IMG_0191

## Tickets

https://trello.com/c/lA1qQ4mB


mhayes853:
Updated the tint color to be green instead of blue based on last night's feedback.
Simulator Screenshot - iPhone 16 Pro - 2024-10-09 at 12 14 20

seanim0920 (approved):

PR #409, opened 2024-10-15
Common Footer View

mhayes853:
Adds 2 new views:
1. `TiFFooterView` which is a footer that we can use for forms and the tab bar for the updated explore screen design.
2. `TiFFormScrollableLayoutView` which can be used to easily create scrollable screens that render a footer at the bottom.

Here's a simple look at the new footer on the edit event form. The background color is now the card color to help make the CTA standout a bit more.
IMG_0195


mhayes853:
Increased the top padding.
IMG_0196

PR #413, opened 2024-10-23
Interactive Narrative Prototype + Audio and haptic utils

seanim0920:
Storybook for the interactive narrative

https://github.com/user-attachments/assets/59a259c1-86c2-4acb-a62e-a03e00da93e1

Added an Audio lib folder for interfacing with expo-av
Modified haptics module to accept custom patterns
Other smaller changes in the comments

https://trello.com/c/u74wFpA8/805-seans-prototype


tifapp:
Linked Ticket: https://trello.com/c/u74wFpA8

lib/Audio.ts
mhayes853, line 1: Should mention that CoreHaptics also has capabilities to play sound effects alongside haptics in the same AHAP pattern. Those sound effects need to be registered with that framework directly, so I'm not so sure if we should commit to these library functions for sound. 

Additionally, I wouldn't mind handling sound effects in the native code within the AHAP interpreter on Android.

For now, I think we can make the code in this file specific to the prototype instead of it being a shared lib in the entire app.

mhayes853 (commented):
Nice prototype, I think this is going in a good direction regarding the journey. I won't comment too much on the quality of the prototype code due to it being a "prototype". However, I think we need to work on tidying up the haptics code since haptics are an extremely core component of the journey. I would imagine this tidying is enough work to be split off into multiple separate PRs.

**Additionally, I want you to place all the code for the prototype inside the same storybook story, and also ensure that all prototype code is within the same directory. This will make it easier to manage our use of storybook for varying prototype/test/preview targets.**

seanim0920:
It was easier to make the prototype using separate storybook screens, and I'm sure it'll be easier for others too. I think instead of putting everything in the same storybook story I'll add a separate category for the prototype-specific stories.

seanim0920:
I agree the haptic code needs more work, although I want to make it available for people to play with. I'll try and keep the code isolated from the lib folder aside from the native module changes.

mhayes853:
Thanks for addressing the multiple storybook issue part. I agree that it's useful, but I would like to have some visual system to ensure that we can decipher prototype stories from main app stories. Additionally, I would like you to nest all the folders that make up your prototype inside another folder. This way, we can use the `Prototypes` folder as a top level folder to store multiple prototypes.

mhayes853:
On the subject of haptics, I would rather we make a few separate PRs to standardize its API so that we can share the code between my upcoming prototype, your prototype, and the main app.

PR #414, opened 2024-10-23
Remove fallback npm cache

seanim0920:
+ restore-keys causes the github action to fallback to an older version of the node modules cache if there's a miss. Since we always want to use the latest dependencies for our project, restore-keys isn't actually needed, so we can remove those. 
+ updated the node modules cache key to include the node version so when we update, it also forces the cache to regen.
+ does not reinstall node modules with each job if it was already cached

https://trello.com/c/j8JodyX0/827-node-update


tifapp:
Linked Ticket: https://trello.com/c/j8JodyX0

mhayes853 (approved):

mhayes853:
Merged with #397.

PR #421, opened 2024-11-12
Android Custom AHAP Interpreter

ZoroarkZark:
## Problem/Solution

This PR creates a custom AHAP pattern interpreter for Android, using the Kotlin language, aiming to allow for Android to use the same patterns as in the iOS implementation for custom haptics. Since Android does not have a set pattern format for haptics, the AHAP pattern format is used as interpreted as in the iOS implementation, to allow for consistency. 

## Implementation Steps

* For this to be performed, an AHAP pattern will be read as a `Map<String, Any?> type`, the type that an object in Javascript/Typescript would be translated as.
* The AHAP pattern is then taken into `createReadablePattern` to translate it into a `HapticPattern` type, which holds the various haptic events in a property `events`, a list of `HapticPatternElements`, by translating, and then sorting, every event in the pattern into the `HapticPatternElement` type with `createReadableEvent`, allowing it to be read by Kotlin code. 
* After this, each event is called one by one by the `HapticPlayer`, through `playHapticPattern`, which would use `playSound` and `playEffect` to do so.

## Tickets

https://trello.com/c/Cq9czN78


tifapp:
Linked Ticket: https://trello.com/c/Cq9czN78

seanim0920 (approved):

PR #422, opened 2024-11-27
API: Update Endpoint Calls

mhayes853:
- Updates existing endpoint calls to use the latest `TiFAPI` schema updates.
  - For now, I've put a TODO in the Auth since it requires a name. GIven that we'll change auth soon, we can leave a true fix to that in another PR.
- Adds the submit function for the edit event form using the new create and edit event endpoints.

## Tickets

 https://trello.com/c/FYY4HhMq


tifapp:
Linked Ticket: https://trello.com/c/FYY4HhMq

seanim0920 (approved):

PR #423, opened 2024-11-30
Friend Request: Generic Friend Request View

mhayes853:
We'll have multiple UI elements in the app that have the ability to both send and accept friend requests. However, the logic and friend request specific UI for these UI elements will be the same, so I decided to extract that logic and shared friend request UI into a `FriendRequestView` and `useFriendRequest` hook respectively.

## Tickets

https://trello.com/c/hESd3xVW


tifapp:
Linked Ticket: https://trello.com/c/hESd3xVW

seanim0920 (approved):

PR #426, opened 2024-12-02
Event Details: Join + Leave Event Update

mhayes853:
Due to the leave event flow has been simplified for the alpha (ie. no longer requiring the host to pick a new host), I updated the `useLeaveEvent` hook to reflect those changes. Additionally, I also made some minor changes to `useJoinEvent` in order to create the `EventUserAttendanceButton` view.

The new event card design will include a join and leave button, which has the same logic as the button on the details screen. Therefore, I extracted `LeaveEvent`, `JoinEvent`, `EventLocationIdentifier` into the `@events` module, and I created a new `EventUserAttendanceButton` that provides a shared button view for joining and leaving the event. `EventUserAttendanceButton` calls both `useJoinEvent` and `useLeaveEvent` under the hood, and renders the join or leave button based on the user's attendance status. Since the button will be rendered in deeply nested component hierarchies, I decided to pass the data functions (ie. The `joinEvent` and `leaveEvent` API calls) through a context. You can override the context values for the button by using `EventUserAttendanceProvider`. I have yet to provide true default values for the monitor and join event in the context, and that will come in a future PR.

## Tickets

https://trello.com/c/fsmKgJXt
https://trello.com/c/QWlwe5VF


event/JoinEvent.tsx
```diff
@@ -161,18 +184,20 @@ export type UseJoinEventEnvironment = {
...
-export type UseJoinEventPermissionStage = {
+export type UseJoinEventPermission = {
   permissionKind: JoinEventPermissionKind
+  bannerContents: JoinEventPermissionBannerContents
   requestButtonTapped: () => void
   dismissButtonTapped: () => void
 }
 
-export type UseJoinEventStage =
+export type UseJoinEvent =
```
ZoroarkZark, line 197: Good change in name, stage a little unclear at first

event/JoinEvent.tsx
```diff
@@ -218,11 +242,16 @@ export const useJoinEventStages = (
   const hasJoined =
     joinEventMutation.isSuccess && joinEventMutation.data === "success"
   const isSuccess = hasJoined && currentPermission === "done"
+  const reset = joinEventMutation.reset
+
+  useEffect(() => {
```
ZoroarkZark, line 247: What does useEffect() usage do instead of having else if for isSuccess?

event/UserAttendance.tsx
```diff
@@ -0,0 +1,101 @@
...
+} from "./LeaveEvent"
+import { EventID, isAttendingEvent } from "TiFShared/domain-models/Event"
+import { TiFAPI } from "TiFShared/api"
+import { mergeWithPartial } from "TiFShared/lib/Object"
+import { EventRegionMonitor } from "@arrival-tracking"
+import { TrueRegionMonitor } from "@arrival-tracking/region-monitoring/MockRegionMonitors"
+import { ClientSideEvent } from "./ClientSideEvent"
+
+export type EventUserAttendanceContextValues = {
+  monitor: EventRegionMonitor
```
ZoroarkZark, line 27: What does monitor perform here? Little confused

mhayes853: It's needed for `useJoinEvent`

ZoroarkZark (approved):

PR #428, opened 2024-12-03
Organize build scripts into a folder

seanim0920:
Cleans up the root by putting all the eas build config files into a folder.

TASK_UNTRACKED


mhayes853 (approved):
Thanks for organizing this.

PR #429, opened 2024-12-03
Create Story script

seanim0920:
Adds a script `npm create-story <StoryName>` making it quicker for us to make new stories.

TASK_UNTRACKED


mhayes853 (approved):
Thanks for this.

PR #436, opened 2024-12-18
New Profile Screen Component

ZoroarkZark:
## Problem/Purpose

The old profile screen was not using the newer functions, and needed a new format to accomodate them. This PR aims to create a new profile screen UI and new profile screen component, in order to have the screen be adapted for the newest codebase, as the original was made under an old version of the codebase.

## Proposed Solution

In order to make this, two files have been created, alongside test files: `UserProfile.tsx` and `UpcomingEvents.tsx`:
Implementation:
1. Uses `userProfile` to either collect the user from storage (if current user) or to use React Query to collect the user from the server, in the hook `useUserProfile`.
2.  Uses `userUpcomingEvents` to collect a specific user's upcoming events with React Query, in the hook `useUpcomingEvents`.
3. The results of the hooks in 1 and 2 are then used in `userProfileView`, as information for two separate components `UserInfoView` and `UpcomingEventsView`, respectively, to construct the profile screen.

## Results

Screenshot_20241217_144353_FitnessApp
Screenshot_20241217_144240_FitnessApp

## Warnings

At the moment, the friend requesting in the profile screen does not change the status of the user.
Due to using `FlatList` in a `ScrollView` being disallowed, I have used a map to directly render the `EventCard`s  into the `BaseUpcomingEventsView`, a core component to the `UpcomingEventsView` . This is less performant, but works for the small scale of the alpha; this would likely need to be changed later.

## Tickets

https://trello.com/c/LBehUsWt


user-profile-boundary/UserProfile.tsx
```diff
@@ -0,0 +1,249 @@
...
+      user: BasicUser
+    }
+  | { status: "blocked-you" | "user-not-found" }
+
+export const userProfile = async (
+  userID: UserID,
+  api: TiFAPI = TiFAPI.productionInstance,
+  storage: AlphaUserStorage = AlphaUserStorage.default
+): Promise<UserProfileResult> => {
+  if (storage) {
```
ZoroarkZark, line 40: Redundant if check

user-profile-boundary/UpcomingEvents.tsx
```diff
@@ -0,0 +1,70 @@
...
+  style
+}: BaseUpcomingEventsProps) => {
+  return (
+    <View style={style}>
+      <Subtitle>{"Upcoming Events"}</Subtitle>
+      {events.map((item, index) => (
+        <View key={index} style={{ paddingBottom: 16 }}>
+          <EventCard event={item} />
+        </View>
+      ))}
```
ZoroarkZark, line 67: Have Flatlist here, structure using the ListHeaderComponent property

user-profile-boundary/UserProfile.tsx
```diff
@@ -0,0 +1,249 @@
...
+}: UserProfileProviderProps) => (
+  <UserProfileContext.Provider value={values}>
+    {children}
+  </UserProfileContext.Provider>
+)
+
+export type UserProfileProps = {
+  userId: UserID
+  userInfoState: ReturnType<typeof useUserProfile>
+  friendRequestState: ReturnType<typeof useFriendRequest>
```
ZoroarkZark, line 115: Call this in the component, don't need as a prop

user-profile-boundary/UserProfile.tsx
```diff
@@ -0,0 +1,231 @@
...
+export const UserProfileView = ({
+  userInfoState,
+  onRelationStatusChanged,
+  upcomingEventsState,
+  style
+}: UserProfileProps) => {
+  return (
+    <View style={style}>
+      <FlatList
+        style={[style, { padding: 16 }]}
```
ZoroarkZark, line 125: ListSeperatorView for placing padding instead

user-profile-boundary/UserProfile.tsx
```diff
@@ -0,0 +1,231 @@
...
+        data={
+          upcomingEventsState.status === "success" &&
+          upcomingEventsState.data.status === "success"
+            ? upcomingEventsState.data.events
+            : undefined
+        }
+        renderItem={({ item, index }) => (
```
ZoroarkZark, line 132: memoizing EventCard for better performance

event/EventCard.tsx
```diff
@@ -32,7 +25,7 @@ export type EventCardProps = {
   style?: StyleProp<ViewStyle>
 }
 
-export const EventCard = ({
+export const EventCard = memo(function EventCard({
```
mhayes853, line 28: Minor Nit: When dealing with a larger component (>50 LOC), it's a bit more readable (removes a level of nesting) to keep the main component internal to the file, and then to export a call to memo on the internal component.

So:
```ts
const _EventCard = (props: EventCardProps) => {
  // Component Code...
}

export const EventCard = memo(_EventCard)
```

mhayes853 (approved):
I'll handle making the visuals better at a later date, but this seems to be good otherwise.

PR #437, opened 2024-12-18
In App Navigation

mhayes853:
Wires all the screens up with React Navigation V7, and makes some usability improvements to the current UI.

Some architectural changes were needed to make this possible.
- Introduces a new `featureContext` function that automatically creates a context with a provider component and hook to consume it. This removes a lot of boilerplate from feature specific contexts.
- Organizes the Location Search screen into a singular entry point component.
- Introduces a `useCoreNavigation` hook that returns an object of functions that navigate to core app screens from components deeply nested within the component hierarchy.
- Introduces a `withAlphaRegistration` HOC that displays the register screen if the user is not registered.

## Tickets

https://trello.com/c/uVjEOxib


tifapp:
Linked Ticket: https://trello.com/c/uVjEOxib

seanim0920 (approved):

.storybook/components/Explore/Explore.stories.tsx
```diff
@@ -22,23 +22,41 @@ export default ExploreEventsMeta
...
-        <NavigationContainer onStateChange={console.log}>
-          <Stack.Navigator screenOptions={{ ...BASE_HEADER_SCREEN_OPTIONS }}>
-            {screens}
-          </Stack.Navigator>
-        </NavigationContainer>
-      </SafeAreaProvider>
-    </TiFQueryClientProvider>
-  </MenuProvider>
+  <AlphaUserSessionProvider storage={storage}>
+    <GestureHandlerRootView>
```
ZoroarkZark, line 34: Was the GestureHandlerRootView included to help fix issues with the gestures in the app, which were mentioned earlier? 
Those issues being:

- The Journey screen being held when too many gestures are being made
- Not being able to swipe to the previous/next screens, while the map is available

.storybook/components/LocationSearch/LocationSearch.stories.tsx
```diff
@@ -63,7 +63,7 @@ const LocationSearchHeader = () => {
 }
 
 const LocationSearchScreen = () => {
-  const picker = useLocationSearchPicker({
+  const picker = useLocationsSearch({
```
ZoroarkZark, line 66: What encouraged the name change? Both to me sound alright

AppEntry.js
```diff
@@ -2,6 +2,8 @@ import { BUILD_TYPE } from "@env"
 import { registerRootComponent } from "expo"
 import "TiFShared"
 
+console.log(BUILD_TYPE)
```
ZoroarkZark, line 5: Are we going to keep this console.log? XD
I know it works, but I hope this won't be intrusive

components/BottomSheet.tsx
```diff
@@ -21,15 +23,18 @@ import Animated, {
   useAnimatedStyle,
   useSharedValue
 } from "react-native-reanimated"
+import { FullWindowOverlay } from "react-native-screens"
 
 export const TiFBottomSheetProvider = BottomSheetModalProvider
 
 type BaseTiFBottomSheetProps = {
   sizing: TiFBottomSheetSizing
+  overlay?: "on-screen" | "above-screen"
```
ZoroarkZark, line 32: So having this overlay prop is strictly for iOS? I also wasn't fully sure what it did; I'm guessing it's to have it appear above a component?

components/Navigation.tsx
```diff
@@ -15,24 +22,66 @@ export type StackNavigatorType<
...
 
+/**
+ * Returns an object of functions to navigate to essential app screens.
+ */
+export const useCoreNavigation = () => {
+  // TODO: - Actually Navigate
+  const navigation = useNavigation()
+  return {
+    presentEditEvent: (edit: EditEventFormValues, id?: EventID) => {
+      const formValues = toRouteableEditFormValues(edit)
```
ZoroarkZark, line 33: So this takes everything needed for an event (title, description, etc.) and adds a startDateTime?

components/Navigation.tsx
```diff
@@ -15,24 +22,66 @@ export type StackNavigatorType<
...
+      })
+    },
+    pushEventDetails: (
+      id: EventID,
+      method: "navigate" | "replace" = "navigate"
+    ) => {
+      // NB: Typescript doesn't like the subscript for some reason, but navigation should always
+      // have a navigate and replace method.
+      ;(navigation as any)[method]("eventDetails", { id, method })
+    },
```
ZoroarkZark, line 62: Why does the syntax have a ";" before the use of `(navigation as any)`? Is that Prettier doing that because of a weird way the functions are formatted?

components/Navigation.tsx
```diff
@@ -15,24 +22,66 @@ export type StackNavigatorType<
...
  */
 export const BASE_HEADER_SCREEN_OPTIONS = {
-  headerStyle: {
-    shadowColor: "transparent",
-    elevation: 0
-  },
   cardStyle: {
     backgroundColor: "white"
   },
+  headerShadowVisible: false,
```
ZoroarkZark, line 76: So the headers now have a drop shadow? I'm a bit confused at this one

components/form-components/ScrollView.tsx
```diff
@@ -17,6 +17,7 @@ export const TiFFormScrollView = ({
   <KeyboardAwareScrollView
     style={[style, styles.scroll]}
     contentContainerStyle={styles.contentContainer}
+    enableResetScrollToCoords={false}
```
ZoroarkZark, line 20: `enableResetScrollToCoords` sounds a little vague; does this mean that if the keyboard is used on the map component, it'd reset the coordinates after use of the keyboard?

core-root/AlphaRegister.tsx
```diff
@@ -101,20 +88,27 @@ export const AlphaRegisterView = ({ state, style }: AlphaRegisterProps) => (
...
     <IfAuthenticated
       thenRender={(session) => <Component {...props} session={session} />}
       elseRender={
-        <AlphaRegisterView state={useAlphaRegister()} style={styles.register} />
+        <SafeAreaView style={styles.register}>
+          <AlphaRegisterView
+            state={useAlphaRegister()}
+            style={styles.register}
+          />
+        </SafeAreaView>
```
ZoroarkZark, line 111: What's the SafeAreaView needed for?

.storybook/components/Explore/Explore.stories.tsx
mhayes853: Components that require gestures need this component since it acts as a context provider for the gesture handler library (similar to how the bottom sheet library needs a `BottomSheetModalProvider`).

AppEntry.js
mhayes853: No, but for some reason it seems that env values won't always reflect the most up to date value from the `.env` file unless we include this log...

components/Navigation.tsx
mhayes853: Yes, it's just that we're not allowed to pass `Date` as a navigation parameter for react-navigation, so we need a function to serialize `startDateTime` to a string such that we don't get the warning message.

components/Navigation.tsx
mhayes853: Yes...

core-root/Home.tsx
```diff
@@ -89,76 +87,65 @@ const TODO = () => (
 )
 
 type FooterProps = {
-  pageIndex: number
   onPageIndexTapped: (index: number) => void
-  onCreateEventTapped: () => void
-  onProfileTapped: () => void
```
ZoroarkZark, line 95: So wait, why were these removed? Did we not need them anymore?

components/Navigation.tsx
mhayes853: This is mainly an iOS problem, where there's a thin outline under the header by default. This prop ensures that shadow is removed.

components/form-components/ScrollView.tsx
mhayes853: This fixes a problem with the edit event form, in which if you dismissed the keyboard, it would scroll to the top of the screen for some reason. Obviously, that was very annoying.

core-root/AlphaRegister.tsx
mhayes853: To ensure that the safe area on devices (and particularly iOS devices) does not impede the UI.

core-root/navigation/EditEvent.tsx
```diff
@@ -0,0 +1,110 @@
...
+    },
+    editEventLocationSearch: {
+      options: { headerShown: false },
+      screen: LocationSearchScreen
+    },
+    ...eventDetailsScreens(),
+    ...profileScreens()
+  }
+})
+
```
ZoroarkZark, line 101: So wait, editEvent and createEvent would use the same screen? Or for, createEvent, is this just a placeholder?

core-root/Home.tsx
mhayes853: `useCoreNavigation`

core-root/navigation/EditEvent.tsx
mhayes853: They use the same screen, but with different screen options (notably the title being different). We could detect event creation vs editing in `editEventForm` and update the header dynamically, but this is annoying to do and it's easier to use 2 routes if there are only 2 static screen option configurations.

components/BottomSheet.tsx
mhayes853: Yes, because on iOS the sheet would appear behind modals presented with the native stack navigator. `"above-screen"` ensures that it always appears above any view.

And it is strictly for iOS, as `FullWindowOverlay` does not work on Android.

.storybook/components/LocationSearch/LocationSearch.stories.tsx
mhayes853: `useLocationsSearch` is more general, also because the picker component is no longer the entry point component for the feature.

core-root/navigation/Root.tsx
```diff
@@ -0,0 +1,50 @@
...
+      screens: { editEvent: EditEventNavigator }
+    }
+  }
+})
+
+const Navigation = createStaticNavigation(Stack)
+
+const linking = { prefixes: ["tifapp://"] }
+
+type RootStackParamList = StaticParamList<typeof Stack>
```
ZoroarkZark, line 38: createStaticNavigation? This is new to me, but when looking this up, apparently it's a different approach to navigation using a "static API". It also highlights deep linking; is that what we're using it for?

edit-event-boundary/FormAtoms.test.ts
```diff
@@ -23,7 +23,7 @@ describe("EventFormValues tests", () => {
       placemark: undefined
     })
     expect(store.get(eventEditAtom)).toEqual({
-      ...DEFAULT_EDIT_EVENT_FORM_VALUES,
+      ..._DEFAULT_FORM_VALUES,
```
ZoroarkZark, line 26: This looks like an update on naming, since there might be another thing to edit, though when I looked up `DEFAULT_FORM_VALUES`, it still primarily looks like it'd be for editEvent.
Where else might we want to use this?

core-root/navigation/Root.tsx
mhayes853: It's just much more convenient than using the "dynamic" API, especially when it comes to the way that we share routes between navigators. We'll also be using it for deeplinking in the near future since it makes that much easier as well.

edit-event-boundary/FormAtoms.test.ts
mhayes853: This is an internal variable for the tests because we now because getting the actual default values for the edit form comes from calling the `defaultEditEventFormValues` function. Using a function allows us to always use the current date when we create the form, rather than only computing the current date once when the app launches.

event/ClientSideEvent.ts
```diff
@@ -57,6 +57,9 @@ export type ClientSideEvent = Reassign<
   ClientSideEventTime
 >
 
+/**
+ * Adds the client received time to an {@link EventResponse}.
+ */
```
ZoroarkZark, line 62: 👍

event/Menu.tsx
```diff
@@ -100,34 +100,40 @@ export const EventActionsMenuView = ({
...
+  return (
+    <MenuView
+      // TODO: - Error UI
+      onPressAction={({ nativeEvent }) => {
+        callbacks[nativeEvent.event as keyof typeof callbacks]()
+      }}
+      actions={formatEventMenuActions(
+        event,
+        EVENT_MENU_ACTIONS_LISTS[state.actionsListKey]
+      )}
```
ZoroarkZark, line 126: The usage of this kind of confuses me:
So from what I can tell, these are set callbacks set as pseudo-enums, and you navigate through them for onPressAction, to pick a callback corresponding to what kind of action is needed?

ZoroarkZark (approved):

PR #440, opened 2025-01-11
Edit Event Performance Improvements

mhayes853:
- Improves view appearance performance on Android by delaying animations that run when the view appears.
- Fixes duration slider trail flickering on iOS.
- Ensures that duration presets with both an hour and minute (eg. "1hr 27m") are split into multiple lines.

TASK_UNTRACKED

### DIFF
--- a/edit-event-boundary/DurationPicker.tsx
+++ b/edit-event-boundary/DurationPicker.tsx
@@ -93,7 +93,7 @@ const BackgroundTrailView = ({
     style={[
       styles.backgroundTrail,
       useAnimatedStyle(() => ({
-        width: sliderPosition.value + INNER_TRACK_GAP * 3,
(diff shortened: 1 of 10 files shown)


edit-event-boundary/DurationPicker.tsx
```diff
@@ -168,10 +172,19 @@ const SliderKnobView = ({
...
-    () => animateToPosition(selectedDimensions),
-    [selectedDimensions, animateToPosition]
-  )
+  const didAppear = useRef(false)
+  useEffect(() => {
+    let timeout: NodeJS.Timeout
+    if (!didAppear.current) {
+      timeout = setTimeout(() => {
+        animateToPosition(selectedDimensions)
+      }, 300)
```
seanim0920, line 181: we should document this somewhere if we need to use timeouts a lot for app animates

seanim0920 (approved):

PR #444, opened 2025-01-13
Explore: Map Panning Improvements

mhayes853:
The explore events map panning had the issue where events on the upper corners of the map would disappear even if it seemed like they should be in view. The reason for this was because we were using too small of a region radius.

The region radius was based on the current region the user has panned to on the map. This region contains the latitude and longitudinal coordinates with an additional 2 properties for the latitude and longitude deltas, respectively. To calculate the radius in meters, we would take the smaller of the latitude and longitude deltas, create a new coordinate based on the minimum delta, and then use the haversine formula to compute the distance between the region coordinate and the new coordinate.

However, using the minimum delta produced a diameter that was only as big as the width of the display on devices with heights ~2x their widths. As the radius only refers to a circular radius, events at the top and bottom of the screen were simply cutoff due to being outside the relatively small circle produced by the radius. Therefore, I decided to use the maximum delta instead, ensuring that the height of the display is used as the diameter, thus producing a far bigger circle.

This region calculation is not finalized, we could try to use a rectangular (or some other polygon) region for instance should we see that it works better. Therefore, I'll leave this in draft until we can agree on a good panning experience.

## Tickets

https://trello.com/c/oRTZtPp9

### DIFF
--- a/api/TiFAPI.ts
+++ b/api/TiFAPI.ts
@@ -60,6 +60,8 @@ declare module "TiFShared/api" {
   }
 }
 
+console.log(API_URL)
+
 TiFAPI.productionInstance = TiFAPIClientCreator(
   awsTiFAPITransport(launchArguments.apiURL ?? new URL(API_URL))
 )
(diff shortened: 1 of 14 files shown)


tifapp:
Linked Ticket: https://trello.com/c/oRTZtPp9

mhayes853:
Will just make the radius really big for now when requesting events. This should work for now.

seanim0920 (approved):

PR #452, opened 2025-01-20
Expandable Map Snippet View

mhayes853:
One of our complaints revolved around not being able to interact with the map snipped in any meaningful way when in the event details view such as panning it or zooming it. Because interacting with a small map snippet isn't very ergonomic from the UX standpoint, I decided to add an expand button that expands the map to fullscreen, and allows interactivity once in full screen.

Here's a look:
https://imgur.com/m1fbhKV

To implement this, I had to make use of `@gorhom/portal` on Android in order to make the map expand to full screen. On iOS, I could use `FullWindowOverlay` to achieve the same effect. This led to some differences on how size and padding were calculated on both OSes.

The animation works by springing the size of the map to full screen using reanimated. This turned out to have acceptable performance even on my ~6 year old S10e, so it should be good to go.

I also adopted this logic into a shared `ExpandableMapSnippetView` used in both the edit event form and event details.

## Tickets

https://trello.com/c/e9PTHy5I

### DIFF
--- a/.storybook/.ondevice/Storybook.tsx
+++ b/.storybook/.ondevice/Storybook.tsx
@@ -1,6 +1,16 @@
 import React, { useState } from "react"
 
-import { FlatList, Text, TouchableOpacity, View } from "react-native"
+import MapSnippetMeta, {
+  Basic as MapSnippet
+} from "../components/MapSnippet/MapSnippet.stories"
+
+import {
+  FlatList,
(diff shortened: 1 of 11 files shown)


tifapp:
Linked Ticket: https://trello.com/c/e9PTHy5I

seanim0920 (approved):

PR #456, opened 2025-01-29
Feedback Screen Into Main App

ZoroarkZark:
## Problem/Purpose

This PR implements the feedback screen into the main navigation system of the app. Currently, the feedback screen does not have full styling.

## Proposed Solution

In order for this to occur, the feedback screen implementation is connected to the `core-root/navigation` section of the codebase, using `helpAndSupportScreens`  as a way to declare the feedback screen in the navigiation..

## Tickets

https://trello.com/c/2gUYMrw1


tifapp:
Linked Ticket: https://trello.com/c/2gUYMrw1

core-root/navigation/Feedback.tsx
```diff
@@ -0,0 +1,30 @@
...
+type HelpAndSupportScreenProps = StaticScreenProps<{
+  user: UserSession
+}>
+
+const HelpAndSupportScreen = ({ route }: HelpAndSupportScreenProps) => {
+  const navigation = useNavigation()
+  useBackButton()
+  const state = useHelpAndSupportSettings({
+    userSession: route.params.user
+  })
```
ZoroarkZark: use withAlphaRegistrationProvider

core-root/navigation/Profile.tsx
```diff
@@ -19,7 +22,21 @@ type ProfileScreenProps = StaticScreenProps<{
...
+  const { userSession } = useUserSession()
   useBackButton()
+  useEffect(() => {
+    navigation.setOptions({
+      headerRight: () => (
+        <TouchableIonicon
+          onPress={async () =>
+            navigation.navigate("helpAndSupport", { user: await userSession() })
+          }
+          icon={{ name: "settings" }}
```
ZoroarkZark: Don't need await function, withAlphaRegistration handles for you

settings-boundary/HelpAndSupport.tsx
```diff
@@ -231,11 +250,12 @@ export const useHelpAndSupportSettings = (
 
 const tryComposeBugReportEmail = async (
   composeEmail: (email: EmailTemplate) => Promise<EmailCompositionResult>,
+  userSession: UserSession,
```
ZoroarkZark: Change userSession to userID, not using other properties in userSession

AppEntry.js
```diff
@@ -2,9 +2,8 @@ import { API_URL, BUILD_TYPE } from "@env"
 import { registerRootComponent } from "expo"
 import "TiFShared"
 
-if (BUILD_TYPE === "storybook") {
-console.log(API_URL)
-console.log(BUILD_TYPE)
+if (BUILD_TYPE !== "storybook") {
+  console.log(BUILD_TYPE)
```
mhayes853, line 6: I don't think this should log be needed anymore.

core-root/navigation/Feedback.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const HelpAndSupportScreen = ({ route }: HelpAndSupportScreenProps) => {
+  const navigation = useNavigation()
+  useBackButton()
+  const state = useHelpAndSupportSettings({
+    userID: route.params.id
+  })
+
+  return <HelpAndSupportView state={state}></HelpAndSupportView>
+}
```
mhayes853:
```suggestion
const HelpAndSupportScreen = withAlphaRegistrationProvider(({ session }: HelpAndSupportScreenProps) => {
  useBackButton()
  const state = useHelpAndSupportSettings({  userID: session.id })
  return <HelpAndSupportView state={state} />
})
```

core-root/navigation/Feedback.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+})
+
+type RoutableHelpAndSupportScreenValues = {
+  id: UserID
+}
+
+type HelpAndSupportScreenProps = WithAlphaRegistrationProps<
+  StaticScreenProps<RoutableHelpAndSupportScreenValues>
+>
```
mhayes853:
```suggestion
type HelpAndSupportScreenProps = WithAlphaRegistrationProps< StaticScreenProps<{}>>
```

core-root/navigation/Profile.tsx
```diff
@@ -14,19 +21,45 @@ export const profileScreens = () => ({
...
+
+const ProfileScreen = withAlphaRegistration(
+  ({ session, route }: ProfileScreenProps) => {
+    const navigation = useNavigation()
+    useBackButton()
+    useEffect(() => {
+      navigation.setOptions({
+        headerRight: () => (
+          <TouchableIonicon
+            onPress={async () =>
```
mhayes853:
```suggestion
            onPress={() =>
```

settings-boundary/HelpAndSupport.tsx
```diff
@@ -30,51 +35,53 @@ const email = (
 })
 
 const emailSection = (name: string) => {
-  return `<strong>${name}</strong><br><br><br>` as EmailSection
+  return `<strong>${name}</strong><br><br><br><br><br><br>` as EmailSection
```
mhayes853: On iOS 3 line breaks is enough, I think we can do a platform check here and only introduce more line breaks for Android.

settings-boundary/HelpAndSupport.tsx
```diff
@@ -30,51 +35,53 @@ const email = (
...
-    [],
-    emailSection(
-      "📝 1. Select one or more feedback topics below and provide the necessary details. (Required)"
-    ),
-    emailSection("a. App functionality (How could the app help you?)"),
-    emailSection(
-      "b. Creative synergy (Are the features of the app helping you progress?)"
+  feedbackSubmitted: (userID?: UserID) =>
+    email(
+      `App Feedback - UserID: ${userID}`,
```
mhayes853: I think I would prefer if the user id was in an attachment. It's too easy for the user to remove the id from the subject line here, especially if they don't like the default subject line and want to change it to help communicate their feedback better.

core-root/navigation/Profile.tsx
```diff
@@ -14,19 +21,45 @@ export const profileScreens = () => ({
...
+
+type ProfileScreenProps = WithAlphaRegistrationProps<
+  StaticScreenProps<RouteableProfileScreenValues>
+>
+
+const ProfileScreen = withAlphaRegistration(
+  ({ session, route }: ProfileScreenProps) => {
+    const navigation = useNavigation()
+    useBackButton()
+    useEffect(() => {
```
mhayes853, line 37: I think it would be best to only show this button if it's the current user profile, and not any random user profile.

core-root/navigation/Profile.tsx
```diff
@@ -14,19 +21,47 @@ export const profileScreens = () => ({
...
+
+type ProfileScreenProps = WithAlphaRegistrationProps<
+  StaticScreenProps<RouteableProfileScreenValues>
+>
+
+const ProfileScreen = withAlphaRegistration(
+  ({ session, route }: ProfileScreenProps) => {
+    const navigation = useNavigation()
+    useBackButton()
+    if (session.id === route.params.id) {
```
mhayes853: This breaks the rules of react, you can't conditionally call a hook like this. You should be getting some warning in the console about the number of hooks being called per render.

lib/UserInfoEmailFile.ts
```diff
@@ -0,0 +1,38 @@
...
+import { UserID } from "TiFShared/domain-models/User"
+import { featureContext } from "./FeatureContext"
+
+export interface UserInfoEmailFile {
+  createTempIDFile: (id: UserID) => Promise<string | undefined>
+  deleteTempIDFile: (id: UserID) => Promise<void>
+}
+
+const createTempUserFile = async (id: UserID) => {
+  const fileURI = cacheDirectory + "userID.txt"
```
mhayes853, line 16: Use the temp directory if it's available in expo-file-system.

lib/UserInfoEmailFile.ts
```diff
@@ -0,0 +1,38 @@
...
+    await writeAsStringAsync(fileURI, content)
+    return fileURI
+  } catch {
+    return undefined
+  }
+}
+
+const deleteTempUserFile = async (id: UserID) => {
+  const fileURI = cacheDirectory + "userID.txt"
+  const content = await readAsStringAsync(fileURI)
```
mhayes853: You don't need to read this, just delete the file.

settings-boundary/HelpAndSupport.tsx
```diff
@@ -30,51 +36,55 @@ const email = (
...
   },
-  questionSubmitted: email(
-    "App Question",
-    [],
-    emailSection(
-      "❓ 1. List your question(s) and provide all relevant details. (Required)"
-    ),
-    emailSection(
-      "📸 2. Provide any supplementary information or screenshots related to these question(s). (Optional)"
+  questionSubmitted: (userID?: UserID, tempUserIDURI?: string) =>
```
mhayes853:
```suggestion
  questionSubmitted: (tempUserIDURI?: string) =>
```

settings-boundary/HelpAndSupport.test.tsx
```diff
@@ -55,37 +63,42 @@ describe("HelpAndSupportSettings tests", () => {
...
       await reportWithoutLogs()
       await waitFor(() =>
         expect(alertPresentationSpy).toHaveBeenPresentedWith(
           HELP_AND_SUPPORT_EMAIL_SUCCESS_ALERTS.reportBug
         )
       )
       expect(composeEmail).toHaveBeenCalledWith(
-        HELP_AND_SUPPORT_EMAILS.bugReported(undefined)
+        HELP_AND_SUPPORT_EMAILS.bugReported(AlphaUserMocks.TheDarkLord.id, [
+          "userURI"
```
mhayes853: Make a constant for this.

settings-boundary/HelpAndSupport.tsx
```diff
@@ -30,51 +36,55 @@ const email = (
...
+        "📸 Provide any supplementary information or files related to the feedback above. (Optional)"
+      )
     ),
-    emailSection("c. Other feedback (Any other general feedback you have)"),
-    emailSection(
-      "📸 2. Provide any supplementary information or files related to the feedback above. (Optional)"
-    )
-  ),
-  bugReported: (compileLogsURI?: string) => {
+  bugReported: (userID?: UserID, attachments?: string[]) => {
```
mhayes853:
```suggestion
  bugReported: (attachments?: string[]) => {
```

settings-boundary/HelpAndSupport.tsx
```diff
@@ -30,51 +36,55 @@ const email = (
...
-  feedbackSubmitted: email(
-    "App Feedback",
-    [],
-    emailSection(
-      "📝 1. Select one or more feedback topics below and provide the necessary details. (Required)"
-    ),
-    emailSection("a. App functionality (How could the app help you?)"),
-    emailSection(
-      "b. Creative synergy (Are the features of the app helping you progress?)"
+  feedbackSubmitted: (userID?: UserID, tempUserIDURI?: string) =>
```
mhayes853:
```suggestion
  feedbackSubmitted: (tempUserIDURI?: string) =>
```

settings-boundary/HelpAndSupport.tsx
```diff
@@ -172,70 +182,104 @@ export const HelpAndSupportView = ({ style, state }: EventSettingsProps) => (
...
+    feedbackSubmitted: async () => {
+      await tryComposeEmail(
+        composeEmail,
+        HELP_AND_SUPPORT_EMAILS.feedbackSubmitted(
+          env.userID,
+          await createTempIDFile(env.userID)
+        ),
         "submitFeedback"
       )
+      await deleteTempIDFile(env.userID)
```
mhayes853: I think I would try to get this call into the `tryComposeEmail` helper.

settings-boundary/HelpAndSupport.tsx
```diff
@@ -172,70 +182,104 @@ export const HelpAndSupportView = ({ style, state }: EventSettingsProps) => (
...
             try {
               await tryComposeBugReportEmail(
-                env.composeEmail,
-                await env.compileLogs()
+                composeEmail,
+                env.userID,
+                [
+                  await createTempIDFile(env.userID),
+                  await compileLogs()
+                ].filter(Boolean) as string[]
```
mhayes853: This filtering work could be performed in `tryComposeBugReportEmail` as it's duplicated quite a bit in the other calls to the function.

lib/UserInfoEmailFile.ts
ZoroarkZark: On the Expo Docs, it listed FileSystem.CacheDirectory as the temp directory, at least from how it worded it:
https://docs.expo.dev/versions/latest/sdk/filesystem/

settings-boundary/HelpAndSupport.tsx
ZoroarkZark: I attempted to move this logic to `tryComposeBugReportEmail`, but this method of filter(Boolean), along with a typecast to a `string[]`, is meant to filter out undefined values returned from `createTempIDFile`, and ignore the value altogether, so it does not get counted in the attachments array; moving it caused the old behaviour of `createTempIDFile` not being properly read due to potentially being `undefined`.

mhayes853: Have `tryComposeBugReportEmail` take a `(string | undefined)[]` as a parameter, and then call the filter on that within the function.

mhayes853 (approved):

PR #461, opened 2025-02-10
Auto-removal of userToken in TiFPreview

ZoroarkZark:
* Implements an interface that deletes the current user token, to prevent a JWT token glitch; said glitch preventing connection to the backend by giving a 403 error.
* Replaces the ephemeral instance in the TiFPreview storybook with the default instance, to allow for testing this new userToken behavior.

## Tickets

https://trello.com/c/0GH5cNnI


tifapp:
Linked Ticket: https://trello.com/c/0GH5cNnI

mhayes853 (approved):
Looks good, we could possibly in a follow up add an explicit action (eg. shaking the device) to reset the token.

PR #462, opened 2025-02-10
Remove attendeeCount from map when attendeeCount at or below 1

ZoroarkZark:
* When AttendeeCount for an event is 1 or below, removes the AttendeeCount from the map marker, on the explore events screen.

## Tickets

https://trello.com/c/UleFqGgP


tifapp:
Linked Ticket: https://trello.com/c/UleFqGgP

mhayes853 (commented):
I think it would be nice if we still kept the badge on the marker, but just removed the attendee count.

mhayes853:
Badge looks a bit weird.

Simulator Screenshot - iPhone 16 Pro - 2025-02-11 at 11 32 46

PR #464, opened 2025-02-11
Live Event Indicators

mhayes853:
Implements an indicator to show the user what events are ongoing, and which events are starting soon.

This is entirely implemented on the client side for now, and it updates the list of live events in real-time. Essentially, components can call the `useLiveEvents` hook to see which events are ongoing and which ones are starting soon for the user. `useLiveEvents` also takes a selector function to scope down on the state needed for the component, as some components will only want to display select parts of the `LiveEvents` type.

`useLiveEvents` is backed by `useSyncExternalStoreWithSelector` and the `LiveEventsStore` class. I decided to write the code that tracks the live events outside of react because it was easier to ensure minimal rerendering within react in that case. For instance, if we tried to write the observing in react, we would likely have to have a `useQuery` somewhere which would rerender all components that used it whenever the query state changed, even if the change was unrelated.

As such, I still relied on the underlying `QueryObserver` that powers `useQuery` within `LiveEventsStore` making the store compatible with react query cache invalidations. Additionally, the store runs a timer every 5 seconds to check if any events are starting that were previously starting soon, and if any events have ended if they were previously ongoing. Whilst we could run a separate timer for each event that emits the update operation when it starts or ends, this would get annoying to manage when query changes occur, and thus I think it's simpler to just run a timer tick every 5 seconds and do a batch check on all the live events.

We can only begin tracking the live events once the user is authenticated. To make this possible, I thought it was best that we added a `beginObserving` method to the `LiveEventsStore`. This method takes a `UserID`, and begins observing the live events for the user based on the methodology I described in the previous paragraph. Additionally, there exists an `observeUserChanges` method on the class, which takes an `AlphaUserStorage` as a parameter, and observes for changes in the authenticated user (this required me to add a new `subscribe` method on `AlphaUserStorage`). When the authenticated user changes, `beginObserving` is called automatically. We call `observeUserChanges` once at the root of the application, and the store will ensure that it is always observing events for the current user.

Lastly, when the user opens the app, they will be presented with a sheet containing their ongoing events, and events starting within the next 2 hours. The explore screen will also always show any ongoing events (but not events that are starting soon) on the bottom sheet regardless of the location the user has panned to on the map.
Simulator Screenshot - iPhone 16 Pro - 2025-02-11 at 12 52 17

## Tickets

https://trello.com/c/Jvikm6aA


tifapp:
Linked Ticket: https://trello.com/c/Jvikm6aA

ZoroarkZark (approved):

PR #475, opened 2025-03-15
Add FullScreenOverlay

ZoroarkZark:
- Adds a basic FullScreenOverlay component for TiFBottomSheet, to handle showing bottom sheet in Explore Events screen.

TASK_UNTRACKED

PR #477, opened 2025-03-19
Fix Modal Presentation

mhayes853:
As seen in #475, the bottom sheet modal would present with a particularly weird gap at the bottom of the screen. This gap was caused by the placements of `TiFBottomSheetModalProvider` within `TiFBottomSheet`. The main reason for the placements of `TiFBottomSheetModalProvider` was to ensure that a sheet marked with the `isTerminallyPresented` prop would not be dismissed when `dismissAll` was called by `useBottomSheetModal` (ie. It wouldn't be dismissed by the automatic modal dismissal mechanism in `useTiFNavigation`).

To preserve the automatic modal dismissal, we can leverage `useFocusEffect` inside `TiFBottomSheet` instead. When the focus effect cleanup handler is ran, we can dismiss the sheet.

## Tickets

https://trello.com/c/3oBkrCgD

### DIFF
--- a/components/Navigation.tsx
+++ b/components/Navigation.tsx
@@ -36,8 +36,7 @@ export type UseNavigationReturn = Omit<
 /**
  * Returns the main navigator in the current navigation context in the app.
  *
- * Use this instead hook of `useNavigation`, as this hook will ensure to dismiss any bottom sheet
- * modals when navigating.
+ * Use this instead hook of `useNavigation` for better type safety.
  */
 export const useTiFNavigation = (): UseNavigationReturn => {
   return useNavigation<UseNavigationReturn>()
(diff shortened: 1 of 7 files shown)


tifapp:
Linked Ticket: https://trello.com/c/3oBkrCgD

components/Navigation.tsx
```diff
@@ -40,21 +40,7 @@ export type UseNavigationReturn = Omit<
...
-    navigation.navigate(...args)
-  }
-  const replace: UseNavigationReturn["replace"] = (
-    ...args: Parameters<UseNavigationReturn["replace"]>
-  ) => {
-    dismissAll()
-    navigation.navigate(...args)
-  }
-  return { ...navigation, navigate, replace }
+  return useNavigation<UseNavigationReturn>()
```
seanim0920, line 42: Update the docstring

seanim0920 (approved):

ZoroarkZark (approved):

PR #479, opened 2025-03-20
Basic Map View Type Picker

mhayes853:
Adds a basic map view type picker based on tonight's discussion. I think we can get more creative with this, but for now it's just a simple menu picker. Android supports 1 additional option over iOS called "terrain". 

Additionally, there's an issue on iOS where the menu doesn't appear when you hit the button in the expanded map view. This is due to `FullWindowOverlay`; we can create a separate ticket to handle this.

IMG_0300
IMG_0299

## Tickets

https://trello.com/c/cbGXJTMT


tifapp:
Linked Ticket: https://trello.com/c/cbGXJTMT

ZoroarkZark (approved):

